PR #1, opened 2016-07-11
Service: update to match runc 1.0

yuqi-zhang:
"runc start" functionality was ported to "runc run" in runc 1.0

### DIFF
--- a/etcd/service.template
+++ b/etcd/service.template
@@ -3,7 +3,7 @@ Description=Etcd Server
 After=network.target
 
 [Service]
-ExecStart=/bin/runc start "$NAME"
+ExecStart=/bin/runc run "$NAME"
 Restart=on-failure
 WorkingDirectory=$DESTDIR
 ExecStop=/bin/runc kill "$NAME"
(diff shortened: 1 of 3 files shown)


giuseppe:
thanks for the patch.  I would like to postpone the merge it until we have runc 1.0 in CHAC.  Is there a way to use the same file both for runc 1.0 and the previous version?

Should we use ExecStartPre and ExecStopPost as well?

https://github.com/opencontainers/runc/issues/903

yuqi-zhang:
I'm not sure about the ExecStartPre and ExecStopPost, but the existing methods do work as it stands. I could take a look at whether these 2 methods affect anything.

As for using the same file, I guess we could put in a check to detect runc version? I suppose there's a quick way to run a small script during ExecStart for that.

giuseppe:
is the container completely destroyed by "kill" or something state is left?

yuqi-zhang:
The container is completely destroyed in my tests so far. The "states" folder for the service is removed.

giuseppe:
hi, could you try if the new version of atomic plus the updates made to the images work with runc 1.0?

Now atomic generates the ExecStart and ExecStop directives, the two updated containers are: docker.io/gscrivano/etcd and docker.io/gscrivano/flannel

yuqi-zhang:
Since everything works with runc 1.0 this PR is no longer relevant. Closing.

PR #2, opened 2016-08-03
add mount for /run/flannel

jasonbrooks:


giuseppe:
Thanks for the patch, I've just merged it

PR #3, opened 2016-08-18
flannel: fix mount for /run/flannel

yuqi-zhang:
Add an extra install script to detect if /run/$NAME exists, where
$NAME is the specified name of the container. If not it will be
created, to prevent runc from complaining.


yuqi-zhang:
Potentially should make a delete script as well? When the container uninstalls a container, /run/$NAME should probably be removed as well?

giuseppe:
Good work!  I am going to merget it.

We could probably delete the directory in the same script, once flanneld exits

PR #7, opened 2016-11-24
[merged] docker: Fix live-restore

giuseppe:
Move shim into a new cgroup so that it won't be killed by runc.


giuseppe:
@rh-atomic-bot r+

PR #14, opened 2016-12-08
bind mount host resolv.conf

jasonbrooks:
needed to access etcd via host name

### DIFF
--- a/flannel/config.json.template
+++ b/flannel/config.json.template
@@ -167,6 +167,7 @@
                         "type": "bind",
                         "source": "/etc/resolv.conf",
                         "options": [
+                                "ro",
                                 "rbind",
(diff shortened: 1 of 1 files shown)


flannel/config.json.template
```diff
@@ -161,7 +161,16 @@
...
                         ]
-                }
+                },
+                {
+                        "destination": "/etc/resolv.conf",
+                        "type": "bind",
+                        "source": "/etc/resolv.conf",
+                        "options": [
+                                "rbind",
+                                "rprivate"
```
cgwalters: This should be `readonly` too right?

giuseppe:
@rh-atomic-bot r+ fd54d68

PR #16, opened 2016-12-12
Docker fix config.json

giuseppe:


rhatdan:
Not sure what you are doing here?

giuseppe:
it seems the problem was using "rbind" instead of "bind" for the mount type.  While it is accepted and everything runs correctly, it is confusing for runc to understand /dev is coming from the host and so to not touch /dev/ptmx.

rhatdan:
We should fix runc to not be specific to bind, but also handle rbind.

rhatdan:
@mrunalp PTAL

rhatdan:
@giuseppe what about the SELinux label changes?

giuseppe:
AFAICS, rbind is an option of the bind type; so it is handled.

giuseppe:
the labels have no effect after we moved to "container_runtime_t" for runc and "container_share_t" for the containers data

mrunalp:
LGTM

rhatdan:
Probably should be 

"selinuxProcessLabel": "system_u:system_r:container_runtime_t:s0" 

But with this label, you are not seeing docker running as container_runtime_t?  How about if you run in permissive mode?

Are you seeing any AVCs?  ausearch -m avc -ts recent.

Break the pull request in two without the SELinux stuff and then you can merge. I am thinking something else is preventing the container_runtime_t from working.

giuseppe:
I've moved the SELinux patches here: https://github.com/projectatomic/atomic-system-containers/pull/17

rhatdan:
@rh-atomic-bot r+

rhatdan:
Oops that does not work here.
LGTM

giuseppe:
we have homu integration for this project, not idea why it didn't get your r+

cgwalters:
rhatdan is now an approver here.  See also https://github.com/servo/homu/issues/34

rhatdan:
Ok sorry about that.

PR #17, opened 2016-12-13
docker: remove selinux label

giuseppe:


rhatdan:
When you run a system container with these labels, are you seeing any errors or AVCS?

giuseppe:
@rhatdan to reply to your previous comment on https://github.com/projectatomic/atomic-system-containers/pull/16:

I don't see any AVCs, and it makes no difference if I run in permissive mode.

I think the config.json file is correct as if I `chcon -R system_u:object_r:bin_t:s0 /var/lib/containers/atomic`, then the process inside runc runs as `system_u:system_r:container_runtime_t:s0`, but as soon as I relabel things as:

```
chcon -R system_u:object_r:container_share_t:s0 /var/lib/containers/atomic
chcon system_u:object_r:container_runtime_exec_t:s0 /usr/bin/runc
```

then it automatically gets: `system_u:system_r:spc_t:s0`

I am using Rawhide with the last version of `gscrivano/docker-centos`.  The only manual change was the relabelling of /usr/bin/runc and /var/lib/containers/atomic

PR #18, opened 2016-12-13
docker-centos: change propagation to shared

giuseppe:
Or "-v FROM:TO:shared" would not work correctly.

### DIFF
--- a/docker-centos/config.json.template
+++ b/docker-centos/config.json.template
@@ -109,7 +109,7 @@
 	    "destination": "/var/lib",
 	    "options": [
 		"rbind",
-		"shared",
+		"rshared",
 		"rw",
 		"mode=755"
 	    ]
@@ -160,7 +160,7 @@
     ],
     "hooks": {},
     "linux": {
-	"rootfsPropagation" : "rslave",
+	"rootfsPropagation" : "shared",
 	"resources": {
 	    "devices": [
 		{


rhvgoyal:
Can you give more details that what is broken and how this patch fixes it. In general, if you make defaults to shared for anything, its a little dangerous proposition.

Also for centos, we run docker with MountFlags=slave and that means that shared does not work anyway.

giuseppe:
I've tried with a simple container: `docker -ti --privileged --rm -v /var/lib/tmp:/a:rw,rshared busybox sh`

and from the container mount a tmpfs on `/a/tmp`.

More than trying to solve this simple case, I thought of privileged Docker containers that could be used to mount on the host file system, even though a system/runc container would probably be more appropriate at this point.

Also, it is a partial solution, since it is possible to mount only the mountpoints that are exported from the host.

If it is a case we won't support, I am fine to drop this patch

rhvgoyal:
I am sorry I still don't understand what's the problem. `shared` should work as long as source directory is outside /var/lib/docker/devicemapper/.

So are you saying that with current docker, -v source:dest:shared not working and a mount created inside container does not propagate to host?

giuseppe:
in the case of running docker in runc, where docker is running in a different mount namespace than the host, "shared" makes a difference, as the mount point will be propagated to the host, while if I don't do that, it will be visible only to docker.  I am using MountFlags=shared in the systemd unit file.

rhatdan:
@rhvgoyal this is for docker running inside of a container.

rhvgoyal:
@rhatdan Sure, but there needs to be a technical explanation of what's the problem and how this patch is fixing it. Converting anything to `shared` opens lots of issues in general.

All I can see is that this patch modifies the spec file passed to runc. So this seems to be equivalent of MountFlags= option of systemd unit file which determines if docker runs in a separate mount namespace and what kind of namespace.

With runc, new container always runs in a new mount namespace. Not sure how do we gain by converting rootfsPropagatio=shared instead of rslave.

rhatdan:
Yes that does not make sense.

giuseppe:
I've just tried with docker-1.10.3-46.el7.centos.14.x86_64 (running on the host, not in a container), and this is what I get:

```
[root@localhost ~]# docker run -ti --privileged --rm -v /home:/a:rw,shared busybox sh
docker: Error response from daemon: Cannot start container c8a9b7b20fb91c7c70d2c9d9b4b1e65b646107598670903526dd8cfc9a3aa36f: Path /home is mounted on / but it is not a shared mount..
```

Just want to make sure this is the behaviour we expect.

rhvgoyal:
It is complaining that / is not shared. If you want a volume to be shared that source mount has to be "shared".

And I suspect it is not shared because centos must be running docker daemon in a slave mount namespace. Both rhel and centos run docker dameon with MountFlags=slave because we have issues
w.r.t container deletion when directory is mounted on in some other namespace.

You can use nsenter to enter docker dameon mount namespace and run `findmnt -o TARGET,PROPAGTION /` and that should shjow you the propagation properties of mount point. 

To get rid of this error, you will have to change property of / mount point in docker mount namespace.

`mount --make-shared /`

And run same command again and it should work this time.

giuseppe:
yes, the reason of that failure is that `"MountFlags=slave"` is in place, and it is the same reason why the system container is failing when `"rootfsPropagation" : "rslave"` is used.

Anyway, since the system container version doesn't work differently than the host version, I am going to close this.

giuseppe:
`mount --make-shared /` in the Docker namespace won't be sufficient though to fix the first issue as it is a different peer group, right?   I am still not able to create a mount from a container and see it from the host, even if docker doesn't complain now about the `:shared` option.

Is the problem of deletion of containers affecting Fedora as well?

rhvgoyal:
container mount point will be visible in docker mount namespace but will not propagate to host as docker itself is running in slave mount namespace.

rhvgoyal:
so on centos, first thing will be to be able to run docker daemon on host and that requires certain functionality from kernel which is very tricky to backport. So it might not happen anytime soon.

giuseppe:
thanks for the information.  What new functionalities are required from the kernel?  I ask since for the Fedora version of this container, for enabling the "live restore" functionality, "rslave" won't probably be enough

rhvgoyal:
if a directory is mounted on in any of the mount namespace, then one can not delete that directory in rhel. While in fedora and upstream kernel it is allowed. 

I think whatever issue you are facing, lets try to fix it in fedora and have a conversation what's the issue and what modifications are required.

rhvgoyal:
We need to do some tests with auto activation of logical volume or volume group disabled. What happens. Will system boot be impacted because local-fs.target will wait for the device to come up which will never come up. If that's the case, I think it is safer to do WantedBy=docker.service. Lets test it first and see what's the impact.

rhatdan:
But then we force a tie to docker.service, when it could be docker-latest.service or ocid.service or docker-centos.service ...

rhvgoyal:
for docker-latest.service, we already have docker-latest-storage-setup, so that's not a problem.

Or how about defining a new target, say container-storage.target. And this mount unit will say WantedBy=container-storage.target and docker.service will say Wants=container-storage.target and After=container-storage.target

That way any runtime can specify dependency on container-storage.target and run after storage has been setup?

rhatdan:
Before we go through all of that lets establish that a failed mount service will block the multi-service-target.

PR #21, opened 2016-12-19
Docker plugins

giuseppe:


docker-fedora/init.sh
```diff
@@ -8,9 +8,15 @@
...
-/usr/libexec/docker/docker-containerd-current \
-    --listen unix:///run/containerd.sock      \
-    --shim /usr/bin/shim.sh &
+# Run all the installed containers
+mkdir -p /run/docker/plugins/
+ls -1 /usr/libexec/docker/*plugin |  \
+while read i;
+do
+    plugin=$(basename $i)
+    test -e /run/docker/plugins/$plugin.sock || mkfifo /run/docker/plugins/$plugin.sock
```
giuseppe: the /usr/libexec/docker/rhel-push-plugin plugin creates it.  I've just checked anyway that all plugins create this socket if it is not present, so I can remove this line from the patch

giuseppe:
pushed some fixup patches ⬆️ 

My tests were to install OpenShift using this image (that include pull all images) and run some containers

giuseppe:
closing this PR as it is being addressed in https://github.com/projectatomic/atomic-system-containers/pull/38

PR #22, opened 2016-12-23
[merged] Docker on Rawhide: better support for live restore

giuseppe:
The previous version was completely broken in case of an upgrade, which is the main reason of having live restore.

This is still a proof of concept, but now Docker containers upgrades are supported.  If we'll decide to go this way, most of the scripts and the new `mount-in-ns` tool must be cleaned up.

Another solution to investigate is to use a chroot directly into /var/lib/containers/atomic/docker.0/rootfs instead of a container so that we do not have to deal with different namespaces, at the cost of dealing with the new mount points in the host mount namespace and not leak them).  A leaked mount point will block container upgrades since we won't be able to rm -rf its rootfs.

### DIFF
--- a/docker-centos/service.template
+++ b/docker-centos/service.template
@@ -17,7 +17,6 @@ LimitNOFILE=1048576
 LimitNPROC=1048576
 LimitCORE=infinity
 TimeoutStartSec=0
-MountFlags=shared
 
 [Install]
 WantedBy=multi-user.target
--- a/docker-fedora/Dockerfile
+++ b/docker-fedora/Dockerfile
@@ -1,8 +1,9 @@
 FROM fedora:rawhide
 MAINTAINER Giuseppe Scrivano <gscrivan@redhat.com>
 
-RUN dnf install -y docker docker-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz && dnf clean all
-ADD shim.sh init.sh init-wrapper.sh /usr/bin/
+ADD mount-in-ns.c /root
+RUN dnf install -y gcc docker docker-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz inotify-tools && gcc /root/mount-in-ns.c -o /usr/bin/mount-in-ns && mkdir -p /usr/lib/modules && dnf clean all
+ADD shim.sh init.sh init-wrapper.sh setup-keep-alive-process.sh /usr/bin/
 
 # system container
 COPY service.template tmpfiles.template config.json.template /exports/
--- a/docker-fedora/config.json.template
+++ b/docker-fedora/config.json.template
@@ -14,6 +14,7 @@
 	    "/usr/bin/init-wrapper.sh"
 	],
 	"env": [
+	    "DESTDIR=$DESTDIR",
 	    "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
 	    "TERM=xterm",
             "NAME=$NAME"
@@ -91,14 +92,18 @@
 	{
 	    "destination": "/proc",
 	    "type": "proc",
-	    "source": "proc"
+	    "source": "proc",
+            "options": [
+                "rprivate",
+                "mode=755"
+            ]
 	},
 	{
 	    "type": "bind",
 	    "source": "/run",
 	    "destination": "/run",
 	    "options": [
-		"rprivate",
+		"shared",
 		"rbind",
 		"rw",
 		"mode=755"
@@ -109,7 +114,7 @@
 	    "destination": "/tmp",
 	    "type": "bind",
 	    "options": [
-		"rprivate",
+		"shared",
 		"rbind",
 		"rw",
 		"mode=755"
@@ -126,31 +131,22 @@
 	},
 	{
 	    "type": "bind",
-	    "source": "/",
-	    "destination": "/host",
-	    "options": [
-		"rbind",
-		"rshared"
-	    ]
-	},
-	{
-	    "type": "bind",
 	    "source": "/etc",
 	    "destination": "/etc",
 	    "options": [
 		"rbind",
-		"rprivate",
+		"shared",
 		"rw",
 		"mode=755"
 	    ]
 	},
 	{
 	    "type": "bind",
-	    "source": "/lib/modules",
-	    "destination": "/lib/modules",
+	    "source": "/",
+	    "destination": "/host",
 	    "options": [
 		"rbind",
-		"rprivate",
+		"shared",
 		"rw",
 		"mode=755"
 	    ]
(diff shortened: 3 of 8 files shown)


rhatdan:
I think we need to continue running in a different mount namespace on RHEL because of devicemapper issues, until the RHEL kernel gets fixed.  I am not sure that live restore works with traditional rhel docker systems because of this issue.  @rhvgoyal would need to look at what you have done.

giuseppe:
The Docker container runs in its own namespace.  Since /var/lib/docker/devicemapper is remounted as MS_PRIVATE, we will need to join exactly this namespace when the container is restarted.  To do this, I am using a separate process which is maintained alive when the Docker container is stopped.  This process mount namespace is later joined when the Docker container restarts.
To complicate things the container rootfs can be deleted, for example when the Docker container is updated.  This means that before the mount namespace can be joined, we must ensure `/usr` is valid.  This part won't work on RHEL, as we need to delete a directory which is still mounted in a different namespace.
To ensure '/mnt' is valid we will have to mount the new '/usr' in the old kept alive mount namespace.  I am using a simple C program included as part of the container image to achieve this, we will probably need to use this:
https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2015-February/msg00064.html

Shortly, this PR supports these two use cases:

1) the Docker container is restarted, and it can still access/delete containers created previously.
2) the Docker container is updated, the '/usr' coming from the new version is used

Another improvement is to keep alive the process only while the Docker container is running or while there are still Docker containers running.  The current implementation is only looking that `/var/lib/docker/containers` exists.

giuseppe:
since the current Docker container for Fedora is completely broken, what about merging this PR and start from here?

rhatdan:
Lets merge, but i think we should think about removing the mount namespace altogether and not running docker in a runc container at all, but just running it in a systemd chroot.  Would this eliminate the problems we are seeing with mountnamespaces?

rhatdan:
@rh-atomic-bot r+

giuseppe:
yes, using a chroot will solve this problem completely

rhatdan:
Lets go that route then.  I think you need to set the SELinux label to container_runtime_t.

giuseppe:
I've tried to get it running in a chroot.  My idea was to mount -R / $(CHROOT_PATH), make it rslave and mount only `/usr` from the cointainer on top of that (which is more or less what this PR is doing in a separate mount namespace).

Unfortunately docker remount `/var/lib/docker/devicemapper` as MS_PRIVATE, so everything that is mounted under $(CHROOT_PATH)`/var/lib/docker/devicemapper` won't propagate to `/var/lib/docker/devicemapper`

rhatdan:
@rhvgoyal PTAL

rhatdan:
Any reason why we are using any rslave?

If you run docker out of ostree in a chroot, with /dev, /run, /proc, /var, ... mounted into the chroot.

giuseppe:
I've tried with binding '/' to the chroot and '/usr' from the container so that docker will be able to see and run from any directory on the system (except those under /usr).

In any case, mounting `/dev`, `/run`, `/proc`, `/var` won't still work.  Docker does a remount of `/var/lib/docker/devicemapper` to `MS_PRIVATE`.  Anything that will be mounted by Docker under `$(CHROOT)/var/lib/docker/devicemapper` won't be propagated to `/var/lib/docker/devicemapper`. 

Once we restart Docker we will lose these mounts, as we will bind again `/var/lib/docker/devicemapper` and we won't be able to access running containers which are left running by the first Docker instance.

rhatdan:
Wouldn't that be the same issue, whether or not docker was running in a chroot.  IE We have this issue in general with docker on devicemapper?

giuseppe:
yes, we have that problem with Docker in a container as well.  My workaround for it is to leave a process running after the Docker container exits, so the mount namespace is not destroyed when we restart Docker and we can later rejoin it, making sure that `/usr` is coming from the container.

Having MS_PRIVATE for Docker makes sense, so that these mount point are not propagated to the different containers, and it seems to happen with any backend not only devicemapper.

rhatdan:
I was asking if docker running on the host, has this issue?  If not why not?

giuseppe:
no, docker running on the host has not this issue as `/var/lib/docker/devicemapper` is not unmounted once Docker stops.  When Docker restarts looks if there is a mount  `/var/lib/docker/devicemapper` is a mount point, if there is one, then it won't be recreated but simply remounted to `MS_PRIVATE`.

rhatdan:
So docker leaves /var/lib/docker/devicemapper mounted when it runs on the host, but if it runs in a chroot it unmounts it?

rhatdan:
Or systemd does?

giuseppe:
yes, I unmount it.  I am experimenting right now on how to leave it mounted so that it can be accessible when we reboot the service (and also when we upgrade it).  What I am trying to do now is to create the root for the service under `/var/lib/docker-container/rootfs` and chroot there

rhatdan:
I was thinking the chroot would be in the ostree directory.  Perhaps we should discuss this in a bluejeans.

giuseppe:
the problem is that the chroot changes when we do a container upgrade.

On an upgrade the rootfs for the container will change from `/var/lib/containers/atomic/docker.0/rootfs` to `/var/lib/containers/atomic/docker.1/rootfs`

rhatdan:
Ok, now I get that, but isn't /var/lib/docker mounted into both, and that where it looks for the content?

giuseppe:
if we bind mount /var/lib/docker to /var/lib/containers/atomic/docker.0/rootfs/var/lib/docker  When Docker starts (assuming we use devicemapper) we have:

if `/var/lib/docker/devicemapper` is not a mount point, then
 mount --bind `/var/lib/docker/devicemapper` `/var/lib/docker/devicemapper`
fi
mount --make-private `/var/lib/docker/devicemapper` `/var/lib/docker/devicemapper`

any container that is mounted under `/var/lib/docker/devicemapper` by Docker then won't be propagated to any other mount point.

Now I am trying with having a single $(CHROOT)/var/lib/docker that won't change when we restart or update the container, so that Docker will end up reusing the same one as it would do when running on the host.

giuseppe:
I have not got a fully working version using the chroot I was talking about before yet, but so far it looks worse than the solution with the container.  Alternatively, I think we can try these other two options:

1) change somehow Docker to not require MS_PRIVATE
2) Do something similar to what Software Collections do, and rely on PATH+LD_LIBRARY_PATH to execute Docker

giuseppe:
I've played a bit with option "2", this is what I've got so far:

https://github.com/giuseppe/atomic-system-containers/commit/2391fff9040e746a95d6a41dc57b599368f29eb0

I've tried it on F25 and live-restore works there, without requiring the hack used in the container version.

giuseppe:
Alternative solution with chroot: https://github.com/projectatomic/atomic-system-containers/pull/25

PR #24, opened 2017-01-10
Update flannel to use sd-notify

yuqi-zhang:
Remove hacky loop to check for Docker configs, and update to use
sd-notify instead to alert systemd when flannel is ready.

### DIFF
--- a/flannel/flanneld-run.sh
+++ b/flannel/flanneld-run.sh
@@ -7,7 +7,7 @@ echo "EnvironmentFile=-/run/$NAME/docker" >> /etc/systemd/system/docker.service.
 # Ensure this file doesn't already exist.
 rm -f run/flannel/subnet.env
 
-/usr/bin/flanneld &
+NOTIFY_SOCKET=/dev/null /usr/bin/flanneld &
 child=$!
 
 while test \! -e /run/flannel/subnet.env
@@ -16,5 +16,7 @@ do
 done
 
 /usr/libexec/flannel/mk-docker-opts.sh -k DOCKER_NETWORK_OPTIONS -d /run/flannel/docker
+
+systemd-notify --ready
 
 wait $child
--- a/flannel/service.template
+++ b/flannel/service.template
@@ -7,9 +7,10 @@ After=$AFTER
 Before=docker.service
 
 [Service]
+Type=notify
+NotifyAccess=all
 ExecStart=$EXEC_START
 ExecStop=$EXEC_STOP
-ExecStartPost=/usr/bin/sh -c "while test \! -s ${RUN_DIRECTORY}/${NAME}/docker; do sleep 0.1; done"
 ExecStopPost=/bin/rm /etc/systemd/system/docker.service.d/$NAME.conf
 Restart=on-failure
 WorkingDirectory=$DESTDIR


yuqi-zhang:
Note that systemd will still prompt a daemon-reload for docker.

rhatdan:
Nice, LGTM, if it works.  :^)

giuseppe:
great!  The only disadvantage is that we will need runc 1.0.  On my Fedora 25 I still see runc version 0.1.1

rhatdan:
@lsm5 @runcom can we get the version of runc on Fedora 25 updated.

yuqi-zhang:
I believe the functionality was added to runc 0.9.0. I did some testing in f25 and it seemed to work fine with 0.1.1 (I put in a delay and docker waited for the sd-notify --ready). Unless you mean something else @giuseppe

giuseppe:
I cannot get it working on my machine.

I think I am hitting: https://github.com/systemd/systemd/issues/2737

I can confirm that if I add a `sleep()` in systemd/src/notify/notify.c so to keep the process alive, then it works:

```
diff --git a/src/notify/notify.c b/src/notify/notify.c
index 70b6f86..3117277 100644
--- a/src/notify/notify.c
+++ b/src/notify/notify.c
@@ -190,6 +190,7 @@ int main(int argc, char* argv[]) {
         }
 
         r = sd_pid_notify(arg_pid ? arg_pid : getppid(), false, n);
+        sleep(5);
         if (r < 0) {
                 log_error_errno(r, "Failed to notify init system: %m");
                 goto finish;
```

If systemd-notify exits before systemd handles the message I have in the journald:
```
Jan 11 21:28:31 helium systemd[1]: Cannot find unit for notify message of PID 15778.
```

Also https://github.com/shishir-a412ed/runc-notify/blob/master/sd_notify.c works for me, as the programs doesn't exit after the call to `sd_notify`.

This was the reason I added the hacky loop in service.template :(

yuqi-zhang:
Strange, I cannot replicate that locally on f25, will take a deeper look first.

Although locally, for a simple run script I set up, the service seems to notify ready before I hit sd_notify, somewhat of the opposite of what you're seeing. I suppose using the c command may be better if it can avoid the potential race condition. Will take a look.

yuqi-zhang:
I still cannot replicate the systemd issue, but I changed it to use shishir's sd_notify and added the binary. Does this version work for you @giuseppe ?

giuseppe:
@yuqi-zhang might it depend from the retcode of sd-notify?  What happens if it exits with success?  It should do that on signals instead of "return 1" and "return 2".

yuqi-zhang:
Ah right, sorry let me fix that

yuqi-zhang:
I opened a separate issue for the service failure for now.

As for systemd-notify, we currently have 3 ways to include it as per discussion with @giuseppe , either as a binary, as a .c to be built in the container from source, or we can include it as part of atomic itself. What do you think @rhatdan ?

rhatdan:
I think adding this to atomic itself is the sanest thing.  atomic-sdnotify.  I was hoping we could do this in python, but I don's see any common bindings.

lsm5:
@rhatdan runc 1.0.0-rc3 has been in -testing for a week now. Has the CVE-2016-9962 fix too. Everyone, please add karma so we can get it into stable: https://bodhi.fedoraproject.org/updates/FEDORA-2017-0200646669

giuseppe:
before we add another tool, @yuqi-zhang, could you please verify if there is any work in progress in systemd/systemd#2739?  Fixing it in systemd would probably be the best solution and we can use directly systemd-notify

rhatdan:
Looks like this stalled 9 months ago.

yuqi-zhang:
Yeah there doesn't seem to be a fix in the foreseeable future. Interestingly this should really only happen in non-root cases (unless I'm mis-understanding the conversation in the original thread). I still haven't replicated the original issue locally either, so I'm not even sure I should open a new issue in systemd or something along those lines...

giuseppe:
it is a race condition (which I can reproduce almost all the time :)).  Another solution would be to pass the PID used by runc inside the container itself, so that systemd-notify can use this PID instead of letting systemd guessing it and failing when the process is already gone.

rhatdan:
Isn't the way this is supposed to work is that PID 1 of the container is supposed to call sd_notify, which triggeres runc to call sd_notify which passes it on to systemd?

giuseppe:
from what I can see, all that runc does is to setup a bind mount to NOTIFY_SOCKET without intercepting anything.

rhatdan:
Is that a bug?  Does systemd verify that the processes that is notifying the socket is the child of systemd?

giuseppe:
It is controlled via `NotifyAccess`:
```
NotifyAccess=

    Controls access to the service status notification socket, as accessible via the sd_notify(3) call. Takes one of none (the default), main or all. If none, no daemon status updates are accepted from the service processes, all status update messages are ignored. If main, only service updates sent from the main process of the service are accepted. If all, all services updates from all members of the service's control group are accepted. This option should be set to open access to the notification socket when using Type=notify or WatchdogSec= (see above). If those options are used but NotifyAccess= is not configured, it will be implicitly set to main.
```

We can specify to systemd-notify the PID of the process that is ready, but inside the container we will have PID from the new PID namespace so it is not really helpful, we will need to pass this information from outside.

rhatdan:
Well is PID1 of the container the same as runc's pid? runc just execs the pd and disappears, So as long as pid1 send the notify, it should work correctly.

giuseppe:
well the issue with the race condition still exists.  I am able to reproduce here, a `sleep()` in systemd-notify after sd_notify is called makes it work on my machine.
If we don't specify any PID then system probably tries to get the PID of the caller process via SO_PEERCRED, but by the time it does that then the process is already dead.  We cannot specify the PID as the PID namespace is different so telling PID 1 doesn't work either.  We will need to pass this from outside, or get it fixed in systemd.

@yuqi-zhang have you a chance to verify it with systemd guys?

giuseppe:
@yuqi-zhang could you try if it still work with https://github.com/opencontainers/runc/pull/1308 for you?  It solves the issue for me here, let's see if it will be accepted upstream :)

yuqi-zhang:
@giuseppe the patch works for me as well. If it gets accepted I'll revert back to the CLI notify. Thanks for the patch!

rhatdan:
@yuqi-zhang Add a comment on the runc patch please.

giuseppe:
@yuqi-zhang the patch got merged into runc.  We can use systemd-notify now.

yuqi-zhang:
@giuseppe I've reverted it to use the CLI sd-notify. Thanks for the patch in runc!

giuseppe:
the patch LGTM, thanks for working on it.

Let's wait until the patched runc gets into Fedora before we merge this PR

giuseppe:
@rh-atomic-bot r+ 18d69b4

PR #25, opened 2017-01-18
[RFC] docker: use a chroot environment

giuseppe:
skip the container solution altogether so that the same mount
namespace can be used by docker on different reboots.


docker-fedora/set_environment.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
```
rhatdan: setup_mounts.sh?

set_environment to me looks like setenv()

rhatdan:
@rhvgoyal PTAL

docker-fedora/set_environment.sh
giuseppe: what about initialize_chroot and clean_chroot?

docker-fedora/init.sh
```diff
@@ -1,15 +1,13 @@
-#!/bin/bash
+#!/bin/sh
 
 # set storage first
 (
-    . /etc/sysconfig/docker-storage-setup
-    /usr/bin/docker-storage-setup
+        . /etc/sysconfig/docker-storage-setup
```
giuseppe: no not really, I had just indented the whole file after I worked on it, I'll fix that

docker-fedora/set_environment.sh
rhatdan: much better

docker-fedora/init.sh
rhatdan: just want the indent to match all over.

You might want to reorganize this content so all of the docker-* can share the same content.

giuseppe:
@rhatdan I pushed a new version for docker-centos as well.

The only issue I am having now is that I need this selinux change to get it running in a chroot with the right label:

```
module docker-chroot 1.0;

require {
	type container_share_t;
	type container_runtime_t;
	class file entrypoint;
}

#============= container_runtime_t ==============
allow container_runtime_t container_share_t:file entrypoint;
```

There are few things that must be changed, but for now it is probably fine to start playing with it.

I managed to install OpenShift with this version of Docker in a chroot.

giuseppe:
@rhatdan what do you think of this PR?  I've tested it more with OpenShift (again full installation + deployment of one app) and it works, the only issue so far is with the selinux additional rule we need.

rhatdan:
```
sesearch -A -s container_runtime_t -p entrypoint -t container_share_t
Found 1 semantic av rules:
   allow container_runtime_t container_share_t : file { ioctl read write create getattr setattr lock relabelfrom relabelto append unlink link rename execute execute_no_trans entrypoint open } ; 
```

We have this in the latest container-selinux, which @lsm5 is attempting to build.  Should work fine on Fedora now.

rhatdan:
@giuseppe did you make the other changes I suggested?

giuseppe:
@rhatdan yes, I have done the other changes and rebased ⬆️

rhatdan:
LGTM

giuseppe:
@rh-atomic-bot r=rhatdan c2c32df

Issue #28, opened 2017-02-02
flannel container not working w/ vxlan backend


jasonbrooks:
I'm able to get the flannel system container working w/ udp, but not with the vxlan backend. Perhaps we need something added to the config.json.template for this? 

Here's what I'm doing:

node 1:

```
node 1:

# atomic install --system gscrivano/etcd
# atomic install --system gscrivano/flannel
# systemctl start etcd
# runc exec etcd etcdctl set /atomic.io/network/config '{"Network":"10.40.0.0/16", "SubnetLen": 24, "Backend": { "Type": "vxlan" } }'
# systemctl start flannel

node 2:

# atomic install --system --set FLANNELD_ETCD_ENDPOINTS=http://10.10.171.92:2379 gscrivano/flannel
# systemctl start flannel
```

Then I try to ping the flannel0 of node 1 from node 2, or vice versa, and I can't ping. With udp backend, it works.

cc @giuseppe

giuseppe:
Are you using NetworkManager?  I see this error when the NetworkManager service is active:
```
Feb 02 12:22:50 atomic-0011.localdomain NetworkManager[16940]: <error> [1486038170.7360] device (flannel.1): Generated connection does not verify: vxlan.remote: property is missing
```

I've tried with the Flannel 0.7 package, but I get the same error, though Flannel 0.7 requires this patch:

https://github.com/projectatomic/atomic-system-containers/pull/29

If I disable NetworkManager then that error disappears.  I am still not able to get vxlan working properly, even if I run it directly on the host, probably some networking issues on my end.

giuseppe:
Closed by: https://github.com/projectatomic/atomic-system-containers/pull/29

PR #29, opened 2017-02-02
flannel: make /proc/fs and /proc/sys writeable

giuseppe:
Flannel 0.7 seems to write here and would fail if it is read only.

### DIFF
--- a/flannel/config.json.template
+++ b/flannel/config.json.template
@@ -193,9 +193,7 @@
 		"readonlyPaths": [
 			"/proc/asound",
 			"/proc/bus",
-			"/proc/fs",
 			"/proc/irq",
-			"/proc/sys",
 			"/proc/sysrq-trigger"
 		]
 	}


rhatdan:
Does this mean we have to run Flannel with spc_t?  Or does this work with SELinux in enforcing mode and the container running as container_t?

rhatdan:
Anyways either way this 
LGTM

giuseppe:
@rhatdan it works with enforcing mode

I'll delegate this to @jasonbrooks as he reported the issue:

giuseppe:
@jasonbrooks to test this, instead of recreating the image, I'd suggest to manually edit /var/lib/containers/atomic/flannel/config.json, drop these two lines and restart the flannel service

jasonbrooks:
@rh-atomic-bot r+

jasonbrooks:
This fixes #28 for me, thanks!

giuseppe:
@rh-atomic-bot r=jasoonbrooks 082e330

rhatdan:
Could someone check that label flannel is running with?

ps -eZ | grep flannel

jasonbrooks:
# ps -eZ | grep flannel
system_u:system_r:container_runtime_t:s0 2151 ? 00:00:00 flanneld-run.sh
system_u:system_r:container_runtime_t:s0 2186 ? 00:00:00 flanneld

rhatdan:
Ok, that is the wrong label. Most likely these should run as spc_t, or something different if we want to confine them.  These are not container_runtimes.  We could run them as container_t, if they can be confined.  Although we would need to ignore MCS Separation.

PR #30, opened 2017-02-13
Add labels to etcd and flannel

yuqi-zhang:
Add initial labels to etcd and flannel with reference to the fedora
registry guidelines.

### DIFF
--- a/etcd/Dockerfile
+++ b/etcd/Dockerfile
@@ -2,6 +2,14 @@ FROM fedora
 
 MAINTAINER Giuseppe Scrivano <gscrivan@redhat.com>
 
+ENV VERSION=0.1 RELEASE=1 ARCH=x86_64
+LABEL BZComponent="etcd-docker" \
+      Name="$FGC/etcd" \
+      Version="$VERSION" \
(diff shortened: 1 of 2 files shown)


yuqi-zhang:
As per discussion with @dustymabe @giuseppe @maxamillion we will try to move this into the fedora container registry, so users can pull/test/use from an "official" repo.

This is a tentative initial label, notably that the bugzilla component is "etcd-docker" and "flannel-docker", same as their rhel counterparts.

ashcrow:
Here is the [documentation reference](https://fedoraproject.org/wiki/Container:Guidelines#LABELS) for those interested.

giuseppe:
Great work!  @ashcrow, thanks for the review :)

@rh-atomic-bot r=ashcrow bda79c3

rhatdan:
@giuseppe did you add new LABEL to allow atomic command to automatically set the --system flag?

giuseppe:
I opened a PR to add it for flannel:

https://github.com/projectatomic/atomic-system-containers/pull/31

rhatdan:
Thanks

PR #35, opened 2017-02-28
Initial Azure agent check-in

baude:


azure/exports/service.template
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=WALA Agent
+Wants=network-online.target sshd.service sshd-keygen.service
+After=network-online.target
+
+[Service]
+ExecStart=/usr/bin/python  -u /usr/sbin/waagent -daemon
+Restart=always
+RestartSec=5
+WorkingDirectory=${DESTDIR/rootfs
```
cgwalters: `}` ?

azure/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM centos:latest
+RUN yum -y install WALinuxAgent && yum -y clean all
+COPY exports /
```
cgwalters: Missing?

baude: I dont understand your comment.  Can you elaborate ?

yuqi-zhang: The current syntax is copying a file called exports into the container's root, which you don't have included. I think for the time being what you want is `COPY service.template /exports/`

azure/exports/service.template
baude: Fixed

azure/Dockerfile
baude: @cgwalters ok now ?

giuseppe:
@yuqi-zhang, @cgwalters does the new version look fine now?

yuqi-zhang:
Files seem fine to me, but just the Dockerfile and the service file is not enough to get it working, so I'm not sure if we should just merge as is as a "check-in" or leave it open until its ready?

giuseppe:
@yuqi-zhang it doesn't use runc.  It spawns a python program so the config.json.template file is not necessary

giuseppe:
@rh-atomic-bot r+ 3d9116d

yuqi-zhang:
@giuseppe I see, just that I was testing on centosAH and it error'ed with  `/usr/bin/python: can't open file '/usr/sbin/waagent': [Errno 2] No such file or directory`, so I was wondering if it was missing some bindings or tmpfiles.template

giuseppe:
@baude, should it be '-u usr/sbin/waagent'? /usr is still the one from the host. Alternatively it can be $DESTDIR/rootfs/usr/sbin/waagent'

giuseppe:
@baude are you still on it or would you prefer if I open a PR for the issue up here?

PR #36, opened 2017-03-01
oVirt Guest Agent System container

vinzenz:
This is the initial implementation for oVirt Guest Agent as a system container for CentOS and Fedora

The agent RPM for CentOS and Fedora can differ in content and therefore are there two different implementations - However the only difference _here_ is the Dockerfile where there's a different FROM and a slightly different description


giuseppe:
bot, add author to whitelist

ovirt-guest-agent-centos/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+      license="ASL 2.0" \
+      architecture="x86_64"
+
+ADD logger_conf /root/logger_conf
+
+RUN dnf -y install ovirt-guest-agent-common
+RUN cat /root/logger_conf >> /etc/ovirt-guest-agent.conf && rm /root/logger_conf
+
+COPY ovirt-guest-agent-install.sh /usr/bin/install.sh
+COPY ovirt-guest-agent-uninstall.sh /usr/bin/uinstall.sh
```
giuseppe: should the destination be `/usr/bin/uninstall.sh`?

I suggest putting these together in a single:

COPY ovirt-guest-agent-install.sh ovirt-guest-agent-uninstall.sh /usr/bin/

ovirt-guest-agent-centos/prep.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+/bin/rm -f /usr/share/ovirt-guest-agent/LockActiveSession.py{c,o}
+/bin/rm -f /usr/share/ovirt-guest-agent/LogoutActiveUser.py{c,o}
+
+# Fake being the host - To show the Atomic Host release
+/bin/rm -f /etc/redhat-release
+/bin/ln -sf /host/etc/redhat-release /etc/redhat-release
+
```
giuseppe: could this be bind mount from the host in the config.json file?  Otherwise it will keep the redhat-release of the build host

ovirt-guest-agent-fedora/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+      license="ASL 2.0" \
+      architecture="x86_64"
+
+ADD logger_conf /root/logger_conf
+
+RUN dnf -y install ovirt-guest-agent-common
+RUN cat /root/logger_conf >> /etc/ovirt-guest-agent.conf && rm /root/logger_conf
+
+COPY ovirt-guest-agent-install.sh /usr/bin/install.sh
+COPY ovirt-guest-agent-uninstall.sh /usr/bin/uinstall.sh
```
giuseppe: same typo here

ovirt-guest-agent-fedora/prep.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+/bin/rm -f /usr/share/ovirt-guest-agent/LockActiveSession.py{c,o}
+/bin/rm -f /usr/share/ovirt-guest-agent/LogoutActiveUser.py{c,o}
+
+# Fake being the host - To show the Atomic Host release
+/bin/rm -f /etc/redhat-release
+/bin/ln -sf /host/etc/redhat-release /etc/redhat-release
+
```
giuseppe: same here, it is probably better to bind mount it from the host

ovirt-guest-agent-fedora/Dockerfile
vinzenz: Fixed thanks!

ovirt-guest-agent-fedora/prep.sh
vinzenz: Done

ovirt-guest-agent-centos/prep.sh
vinzenz: Done

vinzenz:
Actually, uninstall.sh isn't used anymore and needs to be dropped, that's what I just figured out :-) 
That was a relict of the Docker based images where we had the UNINSTALL label

giuseppe:
@rh-atomic-bot r+ 4129416

ovirt-guest-agent-fedora/ovirt-container-hibernate.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+/usr/bin/nsenter --mount=/proc/1/ns/mnt -- /usr/share/ovirt-guest-agent/hibernate $@
```
rhatdan: Doesn't this mean that the /usr/share/overr-guest-agent/hibernate is on the host not in the container?

rhatdan: @giuseppe ^^

giuseppe: ouch I missed this, but the script seems to not be used anywhere.  @vinzenz should we clean this as well?

giuseppe: the way to solve is probably to add this to config.json.template:

"DESTDIR=$DESTDIR",

and then use from the container:

$DESTDIR/rootfs/usr/share/ovirt-guest-agent/hibernate

DESTDIR is replaced at installation time with the path to the container checkout (from the host)

giuseppe: also, why do we need to go into the mount namespace of PID1?

If `hibernate` needs to write to /sys/power/state, we can probably mount `/sys` as rw instead of ro?

vinzenz: I've addressed that in #37 by dropping the wrapper as it is really unnecessary here

PR #37, opened 2017-03-02
ovirt-guest-agent: Drop the hibernate wrapper

vinzenz:
Addresses the comments about the hibernate script from PR #36 after it was merged


giuseppe:
the patch LGTM.  Should we drop the suspend script as well?

vinzenz:
Not sure which you mean :-) hibernate and suspend is done by the same script :-)

giuseppe:
I see there is a ovirt-container-shutdown.sh file left in the container.  Can we drop it as well?

vinzenz:
No we can't otherwise the agent won't be able to reboot the host. 

`# shutdown -h now

Running in chroot, ignoring request.`

giuseppe:
I see.  Thanks for the explanation, I think it is safe to assume /sbin/shutdown is present on the host. Going to merge this:

@rh-atomic-bot r+ 66b05bf

PR #38, opened 2017-03-04
Docker container again

giuseppe:
I've added a workaround for the live-restore issue of Docker having /var/lib/devicemapper/* mounted as MS_PRIVATE.

The trick is to nsenter the shim process namespace, which is the same mount namespace used to run the container initially.  In this way runc is able to access the container rootfs.

Marked as WIP as I've tested only the live-restore functionality.

@rhatdan


giuseppe:
on Rawhide I get this SELinux issue:

```
type=USER_AVC msg=audit(1488758730.150:996): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='avc:  received setenforce notice (enforcing=1)  exe="/usr/lib/systemd/systemd" sauid=0 hostname=? addr=? terminal=?'
type=USER_AVC msg=audit(1488758730.151:997): pid=1332 uid=81 auid=4294967295 ses=4294967295 subj=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 msg='avc:  denied  { send_msg } for msgtype=error error_name=org.freedesktop.DBus.Error.InvalidArgs dest=:1.227 spid=1 tpid=8909 scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:spc_t:s0 tclass=dbus  exe="/usr/bin/dbus-daemon" sauid=81 hostname=? addr=? terminal=?'
```

rhatdan:
Just pushed a fix for container-selinux to allow the spc_t process to dbus chat with the init system.

rhatdan:
What process does the docker daemon and containerd end up running with?

ps -eZ | grep docker

giuseppe:
# ps -eZ | grep docker
system_u:system_r:spc_t:s0       5400 ?        00:00:00 docker-containe
system_u:system_r:container_runtime_t:s0 5535 ? 00:00:06 dockerd-current
system_u:system_r:container_runtime_t:s0 5604 ? 00:00:04 docker-containe

rhatdan:
What is this process? 
system_u:system_r:spc_t:s0 5400 ? 00:00:00 docker-containe

giuseppe:
tht is the shim process to control the container:

root      5400  0.0  0.1 223400  3092 ?        Sl   14:22   0:00 /usr/libexec/docker/docker-containerd-shim-current ad9a9fd09924301fe3468899396e59689aa6c7a35f5470093311ed813c61ec50 /var/run/docker/libcontainerd/ad9a9fd09924301fe3468899396e59689aa6c7a35f5470093311ed813c61ec50 /usr/libexec/docker/docker-runc-current

giuseppe:
there is still an open issue that makes this unusable on Atomic Host which is solved by a newer runC:

https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2017-March/msg00035.html

Until this gets fixed we can keep using the docker-centos container.

I've built both images on Docker Hub as gscrivano/docker-centos and gscrivano/docker-fedora.

```
bash-4.3# rpm-ostree status
State: idle
Deployments:
● fedora-atomic:fedora-atomic/25/x86_64/docker-host
             Version: 25.80 (2017-03-13 23:35:50)
              Commit: 24d4499420ffb2cc49681020bbe5aa6780d780d2b811eab1f5ffea6446b5a4c5
              OSName: fedora-atomic

# atomic pull --storage ostree gscrivano/docker-centos
# systemctl disable --now docker docker-containerd
# atomic install --system --name docker gscrivano/docker-centos
# systemctl start docker
# docker run -d -ti --name foo busybox sleep 10s; machinectl list ; docker rm -f foo
da69947e54aa4d5c0975cb8efb16ad21fbf950cc3ca949698e950dddb0af331c
MACHINE                          CLASS     SERVICE
da69947e54aa4d5c0975cb8efb16ad21 container docker 

1 machines listed.
foo
```

docker-centos/Dockerfile
```diff
@@ -2,11 +2,10 @@ FROM centos
 
 LABEL maintainer="Giuseppe Scrivano <gscrivan@redhat.com>"
 
-RUN yum install -y docker docker-latest container-selinux python-docker-py docker-lvm-plugin docker-rhel-push-plugin docker-novolume-plugin lvm2 iptables procps-ng xz cloud-utils-growpart && yum clean all
+RUN yum install -y docker docker-latest docker-selinux python-docker-py docker-lvm-plugin docker-rhel-push-plugin docker-novolume-plugin lvm2 iptables procps-ng xz cloud-utils-growpart && yum clean all
```
giuseppe: yes, I've added a new patch after the "Revert patch" that updates a few things including the `s|docker-selinux|container-selinux|` package rename

giuseppe:
The limitation of the two upgrades for the container and live-restore has been addressed with this change in the atomic tool:

https://github.com/projectatomic/atomic/pull/944

rhatdan:
One potential problem I see would be if the user has additional content mounted in /, thinks like /opt or /storage which they end up wanting docker to use inside of a container.  We have no way to solve this issue yet.  
Otherwise
LGTM

giuseppe:
@rhatdan yes, I think we can mount the host `/` as rslave in the Docker container and then translate each -v `SRC:DEST` to `/host/SRC:DEST` but this involves adding a wrapper either in the host or in the Docker container itself for converting the requests.
I'd like anyway that the docker executable/wrapper in the host comes from the container itself, let's do this once the RPM feature of system containers is a bit more stable and works on Atomic Host as well.

giuseppe:
@rhatdan can we merge this? :)

rhatdan:
:+1:

giuseppe:
@rh-atomic-bot r+ e9c71e2

PR #40, opened 2017-03-06
Adding open-vm-tools system container for centos and instructions for atomic

dav1x:


yuqi-zhang:
bot, add author to whitelist

open-vm-tools-centos/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+      architecture="x86_64" \
+      BZComponent="open-vm-tools" \
+      maintainer="davis phillips <dphillip@redhat.com>"
+
+ENV SYSTEMD_IGNORE_CHROOT=1
+
+RUN yum -y --disablerepo=\* --enablerepo=rhel-7-server-rpms install yum-utils && \
+  yum-config-manager --disable \* && \
+  yum-config-manager --enable rhel-7-server-rpms && \
+  yum clean all
```
yuqi-zhang: Are these 4 lines necessary? The container seems to build fine without specifically enabling rhel-7-server-rpms, and I get: `Error getting repository data for rhel-7-server-rpms, repository not found` if I leave these in.

open-vm-tools-centos/service.template
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Service for virtual machines hosted on VMware
+Documentation=http://github.com/vmware/open-vm-tools
+ConditionVirtualization=vmware
```
yuqi-zhang: So this container would only work in a vmware virtualized environment? In that case it's probably best to note somewhere in README

yuqi-zhang:
I'm not too familiar with open-vm-tools, but just a couple of things from a system container standpoint:
1. `config.json` should be renamed to `config.json.template`
2. Is it safe to bind so many things as 'rw' in the config? For example, does /etc/shadow have to be rw?

open-vm-tools-centos/Dockerfile
ashcrow: I wouldn't think it would be needed. ```open-vm-tools``` is available in [CentOS 7](http://mirror.centos.org/centos/7/os/x86_64/Packages/). However I wasn't able to find ```open-vm-tools-deploypkg ```.

open-vm-tools-centos/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+
+RUN yum -y install file open-vm-tools perl open-vm-tools-deploypkg net-tools iproute systemd && \ 
+yum clean all 
+
+ADD service.template config.json.template /exports/
```
yuqi-zhang: COPY is preferred over ADD

open-vm-tools-centos/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+ENV SYSTEMD_IGNORE_CHROOT=1
+
+RUN yum -y --disablerepo=\* --enablerepo=rhel-7-server-rpms install yum-utils && \
+  yum-config-manager --disable \* && \
+  yum-config-manager --enable rhel-7-server-rpms && \
+  yum clean all
+
+
+RUN yum -y install file open-vm-tools perl open-vm-tools-deploypkg net-tools iproute systemd && \
```
ashcrow: ```--setopt=tsflags=nodocs``` may be a good addition here unless the documentation for these packages are needed in the system container.

open-vm-tools-centos/Dockerfile
ashcrow: Here is a [reference](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy) to @yuqi-zhang's note.

dav1x:
Will address all suggestions tomorrow. Thanks for the review!

open-vm-tools-centos/Dockerfile
dav1x: This was a hold back from my RHEL build of the container. I've removed.

open-vm-tools-centos/service.template
dav1x: Done.

open-vm-tools-centos/Dockerfile
dav1x: Done

open-vm-tools-centos/Dockerfile
dav1x: Agreed. Good change.

open-vm-tools-centos/Dockerfile
dav1x: @yuqi-zhang unfortunately, open-vm-tools needs rw access to a bunch of files to customize the guest OS on cloning operations. I did change the shadow and passwd to ro as per request

giuseppe:
@dav1x I added a check to the lint script to catch the wrong arch type earlier:

https://github.com/projectatomic/atomic-system-containers/pull/41

yuqi-zhang:
LGTM, thanks for the fixes. What do you think @giuseppe @ashcrow

giuseppe:
it is fine for me as well

@rh-atomic-bot delegate=ashcrow

ashcrow:
@dav1x it might be nice to squash some of the commits but I don't believe it to be required.

@giuseppe leaving merging up to you.

giuseppe:
@rh-atomic-bot r+ ea86431

PR #41, opened 2017-03-07
Add check for supported archs

giuseppe:
the arch accepted in the config.json are the same accepted by golang, so get the list of valid values from golang.

### DIFF
--- a/lint/syscontainers-lint
+++ b/lint/syscontainers-lint
@@ -35,7 +35,8 @@ all_files = [
     "tmpfiles.template"
 ]
 
-found_any_issue = False
+found_any_warning = False
+found_any_error = False
 
 (COLOR_BLACK,
  COLOR_RED,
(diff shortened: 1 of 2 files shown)


lint/syscontainers-lint
```diff
@@ -104,6 +120,11 @@ def check_config_json(path):
         content = f.read()
 
     config = json.loads(content)
+    if config['platform']['os'] not in oses.keys():
+        log(basename, "Invalid os specified: `%s`" % config['platform']['os'], is_error=True)
+    if config['platform']['arch'] not in arches.keys():
+        log(basename, "Invalid arch specified: `%s`" % config['platform']['arch'], is_error=True)
+
```
yuqi-zhang: One minor nitpick, the user can specify a os-arch combo that is invalid, but can pass this test still since the checks are independent. E.g. "freebsd" + "mips64" (Although I doubt this would happen, just for completeness purposes?)

giuseppe:
@yuqi-zhang very good point.  I pushed another version to address your comment

yuqi-zhang:
@rh-atomic-bot r+ 9f432fd

PR #44, opened 2017-03-17
bind mount host resolv.conf

jasonbrooks:
Same idea as #14 -- in order to use the container's etcdctl to configure etcd at a particular hostname, the etcd container needs to be able to resolve hostnames properly.

### DIFF
--- a/etcd/config.json.template
+++ b/etcd/config.json.template
@@ -155,7 +155,17 @@
 				"rw",
 				"mode=755"
 			]
-		}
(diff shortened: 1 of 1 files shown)


giuseppe:
@rh-atomic-bot r+ 34f07d8

@yuqi-zhang just FYI

yuqi-zhang:
Thanks for the patch! Will rebuild etcd in fedora registry.

PR #47, opened 2017-03-24
flannel: make /sys/fs not writeable

giuseppe:
It seems that it is not needed to be writeable as /proc/sys is enough to
get vxlan working.

Reverts part of commit eb5a4794b4c0090402d6d8e804356ace15dcf7cb

### DIFF
--- a/flannel/config.json.template
+++ b/flannel/config.json.template
@@ -203,6 +203,7 @@
 		"readonlyPaths": [
 			"/proc/asound",
 			"/proc/bus",
+			"/proc/fs",
 			"/proc/irq",
 			"/proc/sysrq-trigger"
 		]


giuseppe:
@rh-atomic-bot delegate=aveshagarwal

aveshagarwal:
@giuseppe LGTM. Has anyone tested it on fedora? As I have only tested it on rhel.

giuseppe:
@aveshagarwal. I and @jasonbrooks tested the version with both `/sys/fs` and `/proc/sys` writeable.  This version is not tested, just wanted to make it similar to what we have on RHEL.  I can give it a try this week.

ashcrow:
@giuseppe is this still needed?

giuseppe:
@ashcrow I have not tested it on Fedora but since the change is already used for the RHEL image, I think it should be fine to merge this.

ashcrow:
@rh-atomic-bot r+ c40ce7d

PR #48, opened 2017-03-28
QEMU Guest Agent container

vinzenz:
This is a containerization of the QEMU Guest Agent - It won't have all features working just yet, however the container is running like this already.


giuseppe:
LGTM!

@rh-atomic-bot r+ 07a865e

PR #49, opened 2017-03-28
Docker add AUDIT_READ as well

giuseppe:
I thought `AUDIT_READ` would never be needed by Docker but I forgot of the `--privileged` case.


rhatdan:
LGTM

giuseppe:
@rh-atomic-bot r+ 8a2599a

PR #52, opened 2017-03-29
Revert CAP_AUDIT_READ from Docker Centos

giuseppe:
the cap is not available there so drop it.

Given the differences between Fedora and CentOS/RHEL, from now on we must make clear that docker-centos is based on CentOS and it is the version to use on CentOS/RHEL.

Differently on Fedora docker-fedora must be used.


giuseppe:
@rh-atomic-bot r+ 765fa4f

PR #53, opened 2017-03-30
azure: use $DESTDIR

baude:
azure: use destdir and WALA_ROOT
    
When defining environment variables, use the DESTDIR variable
instead of static, fq-names.  Alse, define WALA_ROOT for use
with the agent when in a system_container.

### DIFF
--- a/azure/exports/service.template
+++ b/azure/exports/service.template
@@ -4,11 +4,13 @@ Wants=network-online.target sshd.service sshd-keygen.service
 After=network-online.target
 
 [Service]
-ExecStart=/usr/bin/python  -u /usr/sbin/waagent -daemon
+Environment=PYTHONPATH=$DESTDIR/rootfs/usr/lib/python2.7/site-packages/
+Environment=WALA_ROOT=$DESTDIR/rootfs
+ExecStart=/usr/bin/python  -u usr/sbin/waagent -daemon
 Restart=always
 RestartSec=5
-WorkingDirectory=${DESTDIR}/rootfs
-RuntimeDirectory=${NAME}
+WorkingDirectory=/var/lib/containers/atomic/wala-container.0/rootfs
 
 [Install]
 WantedBy=multi-user.target
+


azure/exports/service.template
```diff
@@ -4,11 +4,14 @@ Wants=network-online.target sshd.service sshd-keygen.service
...
 [Service]
-ExecStart=/usr/bin/python  -u /usr/sbin/waagent -daemon
+Environment=PYTHONPATH=$DESTDIR/rootfs/usr/lib/python2.7/site-packages/
+Environment=WALA_ROOT=$DESTDIR/rootfs
+ExecStart=/usr/bin/python  -u usr/sbin/waagent -daemon
 Restart=always
 RestartSec=5
-WorkingDirectory=${DESTDIR}/rootfs
-RuntimeDirectory=${NAME}
+WorkingDirectory=/var/lib/containers/atomic/wala-container.0/rootfs
```
giuseppe: shouldn't this still be `$DESTDIR/rootfs`?

azure/exports/service.template
```diff
@@ -4,11 +4,14 @@ Wants=network-online.target sshd.service sshd-keygen.service
...
+RuntimeDirectory=wala-container
```
giuseppe: does the container use the runtime directory at all?  Could we just drop this line?

baude: Sure, I dropped it and updated

giuseppe:
could you use $DESTDIR instead of hardcoding the path?  That will be helpful for atomic updates as the service will point to the correct location on the file system when using a different deployment (ending in .1)

giuseppe:
@baude would you like that I take care of this PR?

azure/exports/service.template
ashcrow: 👍 I believe this should be ``$DESTDIR/rootfs``

ashcrow:
@baude would you like @giuseppe to handle this?

ashcrow:
**bump**

PR #54, opened 2017-04-04
docker-fedora: use runc version spec: 1.0.0-rc5

giuseppe:
Fedora rawhide already has the latest version of runC with the breaking change for the capabilities block, so let's update the configuration to support it.


ashcrow:
👍 I agree.

yuqi-zhang:
I'm not sure if we need all the capabilities for all categories, but that's not a very high priority issue.

@rh-atomic-bot r+ f4344c2

PR #56, opened 2017-04-04
docker-centos: use rhel-push-plugin

giuseppe:

### DIFF
--- a/docker-centos/init.sh
+++ b/docker-centos/init.sh
@@ -29,6 +29,7 @@ do
 done
 
 exec /usr/bin/docker-current daemon \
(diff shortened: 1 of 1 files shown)


giuseppe:
@rh-atomic-bot delegate=edsantiago

edsantiago:
Confirmed that rhel-push-plugin ships in image (`yum install` present in Dockerfile; file present in distributed image). Confirmed that this patch causes `docker push` of a RHEL-based image to fail with the expected '...not allowed to be pushed to docker.io' error. Confirmed via strace that rhel-push-plugin process is being invoked. LGTM.

PR #59, opened 2017-04-25
WIP: Docker system container

ashcrow:
This PR focuses on the ``docker-centos`` image:

* Updates the image to be in line with current linting
* Adds light documentation
* Attempts to increase readability
* Avoid conflicting with package install of docker

Requires: https://github.com/projectatomic/atomic/pull/984


ashcrow:
cc @giuseppe @cooktheryan

docker-centos/Dockerfile
```diff
@@ -1,12 +1,20 @@
...
 
-LABEL maintainer="Giuseppe Scrivano <gscrivan@redhat.com>"
+LABEL maintainer="Giuseppe Scrivano <gscrivan@redhat.com>" \
+      name="docker-centos" \
+      version="0.1" \
+      atomic.type="system" \
+      architecture="x86_64"
 
-RUN yum install -y docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz inotify-tools && mkdir -p /usr/lib/modules && yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-el-candidate/x86_64/os/ && yum install -y --nogpgcheck docker-rhel-push-plugin && yum clean all
+RUN yum install --setopt=tsflags=nodocs -y docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz inotify-tools && \
```
stevekuznetsov: Might want to have the `rpm -V` step after `yum install` to ensure that all packages were installed. `yum install` will exit as successful even if one of the packages in the list could not be installed.

docker-centos/Dockerfile
```diff
@@ -1,12 +1,20 @@
...
+mkdir -p /usr/lib/modules && \
+yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-el-candidate/x86_64/os/ && \
+yum install --setopt=tsflags=nodocs -y --nogpgcheck docker-rhel-push-plugin && \
+yum clean all
 
 ADD init.sh /usr/bin
 
 # system container
-COPY service.template tmpfiles.template config.json.template /exports/
+COPY daemon.json service.template tmpfiles.template config.json.template /exports/
```
giuseppe: Newer versions of atomic can copy arbitrary files to the host.  It is enough to copy the `daemon.json` file under `/exports/hostfs/`. e.g. `/exports/hostfs/etc/foo` is copied to the host `/etc/foo`

We can make the file name depend from the container name with:
```
  "renameFiles" : {
        "/etc/daemon.json" : "/etc/sysconfig/daemon-$NAME.json"
    },
```

It depends if we want to allow multiple versions of the Docker container to be installed at the same time, or have different configurations.  As we could just use a single file, and the container itself could verify if it exists on startup and create the file if it doesn't exist.

ashcrow:
Will update with both suggestions. Thanks @stevekuznetsov and @giuseppe!

giuseppe:
@ashcrow, no not yet, these last features miss a proper documentation

ashcrow:
@giuseppe / @stevekuznetsov updated with your notes.

@giuseppe: I was unable to get the rename to work. PTAL and let me know what I'm missing. I tried both files and directories.

docker-centos/config.json.template
```diff
@@ -362,6 +362,11 @@
 	    ]
 	}
     ],
+    "renameFiles" : {
```
giuseppe: This is a system container feature that should go in a separate file `manifest.json'.
For example we have this `manifest.json` file for the Flannel container: https://github.com/projectatomic/atomic-system-containers/blob/master/flannel/manifest.json

ashcrow: @giuseppe ahh, thanks! I'll update shortly.

ashcrow:
Rebasing ..

docker-centos/config.json.template
```diff
@@ -323,7 +333,7 @@
         },
 	{
 	    "type": "bind",
-	    "source": "${STATE_DIRECTORY}",
+	    "source": "/var/lib/docker-$NAME",
```
giuseppe: what about using `/var/lib/$NAME`?  In this way if we install a container called `docker` we are able to use the existing storage (if any) used by Docker on the host

ashcrow: Sure.

PR #60, opened 2017-04-25
lint: Add fallback for older go tool

ashcrow:
A fallback go tool command is now used as a fallback for those who
are on older golang installs.
```
    $ rpm -q --queryformat=%{VERSION}\\n golang
    1.6.4
    $ ./lint/syscontainers-lint docker-centos/
    Unable to find arch using 'go tool dist list -json'
    config.json.template: found mount point /var/lib.  Use ${STATE_DIRECTORY} instead
    $
```

cc: @yuqi-zhang

### DIFF
--- a/lint/syscontainers-lint
+++ b/lint/syscontainers-lint
@@ -98,7 +98,9 @@ def check_git(path):
                 log(i, "file has not staged changes")
 
 def get_go_os_arch():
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
Makes sense

@rh-atomic-bot r+ 6e4a382

PR #64, opened 2017-05-08
Add container-storage-setup

giuseppe:
See https://bugzilla.redhat.com/show_bug.cgi?id=1448372

Adds ``container-storage-setup`` to containers.

### DIFF
--- a/docker-centos/Dockerfile
+++ b/docker-centos/Dockerfile
@@ -7,8 +7,8 @@ LABEL maintainer="Giuseppe Scrivano <gscrivan@redhat.com>" \
       architecture="x86_64"
 
 
-RUN yum install --setopt=tsflags=nodocs -y docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
-rpm -V docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
+RUN yum install --setopt=tsflags=nodocs -y docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
+rpm -V docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
 mkdir -p /usr/lib/modules && \
 yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-el-candidate/x86_64/os/ && \
 yum install --setopt=tsflags=nodocs -y --nogpgcheck docker-rhel-push-plugin && \
--- a/docker-centos/Dockerfile
+++ b/docker-centos/Dockerfile
@@ -10,7 +10,7 @@ LABEL maintainer="Giuseppe Scrivano <gscrivan@redhat.com>" \
 RUN yum install --setopt=tsflags=nodocs -y docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
 rpm -V docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
 mkdir -p /usr/lib/modules && \
-yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-el-candidate/x86_64/os/ && \
+yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-common-candidate/x86_64/os/ && \
 yum install --setopt=tsflags=nodocs -y --nogpgcheck docker-rhel-push-plugin && \
 rpm -V docker-rhel-push-plugin && \
 mkdir -p /exports/hostfs/etc/docker && \
--- a/docker-fedora/Dockerfile
+++ b/docker-fedora/Dockerfile
@@ -6,8 +6,8 @@ LABEL maintainer="Giuseppe Scrivano <gscrivan@redhat.com>" \
       atomic.type="system" \
       architecture="x86_64"
 
-RUN dnf install --setopt=tsflags=nodocs -y docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz oci-register-machine \
-    && rpm -V docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz oci-register-machine \
+RUN dnf install --setopt=tsflags=nodocs -y docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz oci-register-machine \
+    && rpm -V docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz oci-register-machine \
     && mkdir -p /usr/lib/modules && dnf clean all
 
 RUN ln -s /usr/libexec/docker/docker-runc-current /usr/bin/docker-runc


docker-centos/Dockerfile
```diff
@@ -7,8 +7,8 @@ LABEL maintainer="Giuseppe Scrivano <gscrivan@redhat.com>" \
       architecture="x86_64"
 
 
-RUN yum install --setopt=tsflags=nodocs -y docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
-rpm -V docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
+RUN yum install --setopt=tsflags=nodocs -y docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
+rpm -V docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
 mkdir -p /usr/lib/modules && \
 yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-el-candidate/x86_64/os/ && \
```
lsm5: @giuseppe you should change the koji tag to virt7-docker-common-candidate. I've stopped building packages for -el-candidate, unless of course, someone else is continuing to build here.

Since docker repos were a little different for rhel and fedora, i did request for -el-candidate koji tags in CentOS, but now they're mostly unified, so -common-candidate is where I send builds.

docker-centos/Dockerfile
```diff
@@ -7,10 +7,10 @@ LABEL maintainer="Giuseppe Scrivano <gscrivan@redhat.com>" \
       architecture="x86_64"
 
 
-RUN yum install --setopt=tsflags=nodocs -y docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
-rpm -V docker container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
+RUN yum install --setopt=tsflags=nodocs -y docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
+rpm -V docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin docker-lvm-plugin lvm2 iptables procps-ng xz oci-register-machine && \
 mkdir -p /usr/lib/modules && \
-yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-el-candidate/x86_64/os/ && \
+yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-common-candidate/x86_64/os/ && \
```
lsm5: lgtm

rhvgoyal:
what does this PR do? Some description will help for those who are not involved with systems containers.

ashcrow:
Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1448372

ashcrow:
@rh-atomic-bot r+

rhvgoyal:
Does it work? I mean container-storage-setup uses lvm and lvm developers were not happy with the idea of running lvm from containers. especially with udev running from host.

rhatdan:
This should work as long as you install a OS Specific container.  I don't think we should add container-storage-setup to RHEL or Centos containers yet, since we don't officially support it until RHEL7.4  We want to get some play in Fedora.

As @rhvgoyal is suggesting, their could be issues in using lvm commands within a container that differ from the hosts versions.   We need to make sure that udev works properly when new thin pools and devices are created.

ashcrow:
@rh-atomic-bot r-

ashcrow:
Stopping merge since more conversation has come in.

rhvgoyal:
How system containers handle metadata created by container when system container is removed. Do we remove metadata also from host? IOW, say when docker system container is removed, we remove /var/lib/docker/ as well from host?

container-storage-setup puts its metadata in /var/lib/container-storage-setup. What happens to that data. 

What about lvm?

rhvgoyal:
I am assuming that we leave metadata behind ( and bind mount /var and /etc in container). If we don't cleanup the data from host, then we will run into interesting issue of forward compatibility. Say somebody installs docker-1.12, runs it and later removes docker-1.12 container and installs docker-1.10. And docker-1.10 will not be able to deal with docker-1.12 metadata. This will be true for all the tools which are saving their meatadata on host (as opposed to inside container).

So with these system containers, is metadata specific to container and gets cleaned up with container or is it left behind on host?

rhvgoyal:
What happens when lvm2 on host is newer than one installed in system container. docker makes use of libdevmapper, and that will interact with libudev (inside container), and that might not be compatible with data on host? (This will fall in forward compatibility issue. Older tools capable of understanding new data).

ashcrow:
> So with these system containers, is metadata specific to container and gets cleaned up with container or is it left behind on host?

@giuseppe correct me if I'm wrong here ..

``/var/lib/docker`` is mounted to ``%{STATE_DIRECTORY}`` and noted in the ``tmpfiles.template`` to be created, but is not explicitly set to be preserved at systemd cleaning ... though the age is set to `-` meaning no cleaning will be done.

rhatdan:
Meta Data will be left on the host.  THis is the same problem with atomic host or rolling forward and back on RPMs.  It will be up to the admin to cleanup his metadata.

rhatdan:
@rhvgoyal Correct if we have a mismatch between lvm on the host and in the container we could have problems.  Which is why we want to add the ability to tie the version of the container image to the host using rpm.  But for now, this is a possible issue.

rhvgoyal:
I am afraid that we don't have any mechanism to enforce versioning with these systems containers. Now if software fails in interesting ways, first it will take long time to figure out what went wrong and ultimately it will be pushed back saying that this is configuration issue.

Trying system containers as proof of concept in fedora is fine. But if we are building these for rhel/centos, I feel we should have proper mechanism to be able to address those version incompatibilities issue before we go forward. Otherwise we will have support headaches.

giuseppe:
Correct.  We must ensure the version of the packages in the system container are compatible with the host.  This is why we have one system container for Fedora and one for CentOS.  if we mix versions of the container with the host weird stuffs happen.  For example we won't be even able to read journald logs if the two versions of systemd in the container and in the host are different.

Removing the system container doesn't remove `/var/lib/docker`.

For a better version checking, we already have some basic support for .rpm installation (we are dealing with some issues on Atomic Host now before we can start fully using it) .  The plan is to track the container installation with an .rpm installed on the host, where we can specify the dependencies from the host packages.

ashcrow:
Any reason to not merge this for testing?

rhvgoyal:
@giuseppe If we have only one system container, then container bits can be shipped with host. IOW, usefulness of shipping something in containers will kick in only if we have more than one container and different containers are having different versions of software?

Say we have two system containers. One shipping docker-1.10 and other is having docker-1.12. Say libudev in first container is version X and libudev in second container is version Y. Assume Y > X and udev db  written by version Y can't be read by version X.

Now what kind of dependencies system container will put on host software. I mean if I install docker-1.10 container then its libudev might not even be able to read udev db written by newer libudev on host.

If you specify software version to be same on host and container, then what's the need of shipping system container.

If we say that docker-1.10 container needs libudev version >=X, then version Y will satisfy that requirement and still be incompatible.

Will be good if you can throw some light on the details here.

rhvgoyal:
lets look at systemd example. My understanding is that this is updated reguarly. I think libudev also comes from systemd now. So this will require that systemd version inside container should be same systemd version on host otherwise atomic system container installation should fail. That means whenever a system container is updated, it can be installed only on a specific atomic host which has that version of systemd and it will refuse to start on other atomic host images running a different version of software.

giuseppe:
we could split the container and have another one for the storage setup.  Altough, wouldn't this issue be alleviated with the migration to overlayfs2?

rhvgoyal:
Packaging storage  setup in a separate container will also have fundamental design issue. Moving to overlay2 will work around this particular issue of libudev. But might show up in other form. 

Anyway, I wanted to bring this up for conversation and highlight that we have a potential issue when we are shipping a piece of software both on host as well as in containers and then both piece of software are supposed to work on same data and that can have interesting results.

For now, I will not block you from merging this PR. We will probably have some interesting bugs down the line.

ashcrow:
@rh-atomic-bot r+ be2ba20

rhatdan:
Systemd has been updated ONCE in the run of RHEL7, and they have no plan to update it again.

PR #65, opened 2017-05-08
Remove use of sysconfig docker from containers

ashcrow:
- Move /etc/sysconfig/docker to /etc/sysconfig/docker.orig
- Add a note in /etc/sysconfig/docker.orig pointing to daemon.json
- Remove /etc/sysconfig/docker from systemd templates

Related to https://github.com/openshift/openshift-ansible/pull/4106#discussion_r115111843

### DIFF
--- a/docker-centos/service.template
+++ b/docker-centos/service.template
@@ -3,7 +3,6 @@ Description=Docker service
 After=network.target
 
 [Service]
-EnvironmentFile=-/etc/sysconfig/docker
 EnvironmentFile=-/etc/sysconfig/docker-storage
 EnvironmentFile=-/etc/sysconfig/docker-network
 Environment=GOTRACEBACK=crash
--- a/docker-fedora/service.template
+++ b/docker-fedora/service.template
@@ -3,7 +3,6 @@ Description=Docker service
 After=network.target
 
 [Service]
-EnvironmentFile=-/etc/sysconfig/docker
 EnvironmentFile=-/etc/sysconfig/docker-storage
 EnvironmentFile=-/etc/sysconfig/docker-network
 Environment=GOTRACEBACK=crash


docker-centos/Dockerfile
```diff
@@ -16,6 +16,10 @@ rpm -V docker-rhel-push-plugin && \
 mkdir -p /exports/hostfs/etc/docker && \
 yum clean all
 
+# Move the /etc/sysconfig/docker out of the way as daemon.json is the proper config location
```
giuseppe: do we need to do this?  I am afraid we will break the existing configuration for docker on the host, in case the user wants to still keep it.  Is it enough to drop it from the .service file?

ashcrow: @giuseppe We don't, but we do have to do something to the file. It's part of ``docker-common`` and will lay down otherwise.

rhatdan: We can do this, except we need to be able to handle upgrades.

ashcrow: @giuseppe / @rhatdan would it be easier for upgrading to drop ``/etc/sysconfig/docker`` from ``service`` template?

The problem is if both ``daemon.json`` and ``/etc/sysconfig/docker`` are used, docker fails to start. It notes that it received the same options (even if the values are the exact same).

giuseppe: upgrade should be already managed no?  atomic should not overwrite `daemon.json` when it is already present, and with this change the `/etc/sysconfig/docker` is renamed.  Altough, I think it is enough to drop it from the systemd unit file for it to be ignored from the docker system container at startup

rhatdan:
Yes that works for me. We just need to make sure the daemon.json works properly with the --add-registry and --block-registry calls.

ashcrow:
Amended to only remove ``EnvironmentFile=-/etc/sysconfig/docker``

docker-centos/Dockerfile
ashcrow: I went ahead and changed to just remove it from the ``service.template``.

giuseppe:
LGTM.  If @rhatdan is fine, we can merge this

richm:
openshift origin-aggregated-logging - fluentd running in a container mounts /etc/sysconfig and uses /etc/sysconfig/docker to determine what log driver docker is using.  Will this method still be available?  If not, what is the recommendation for fluentd to determine which log driver to use?

ashcrow:
@richm For system containers the answer would be no, it would not be available. Over time the openshift installer will likely move as well now that ``daemon.json`` is part of the package.  Having data in both the "legacy" ``/etc/sysconfig/docker`` and the "current" ```/etc/docker/daemon.json``` causes docker to error on start up (even if the values are exactly the same). I'd recommend the following:

- Look for the value in ```/etc/docker/daemon.json```
- If the above fails, fall back trying ``/etc/sysconfig/docker``

After a release or two you will likely be able to remove the fallback.

richm:
@ashcrow is there perhaps a better way (more portable, fewer volume mounts, etc.) for the fluentd container to determine which log driver docker is using?

ashcrow:
@richm The following may work depending on what you are looking for:

```
$ sudo docker info | grep -i log
Logging Driver: journald
```


Using a ```cut -d ":" -f 2``` or something similar if needed.

richm:
@ashcrow I need something that works inside of the fluentd docker container.

rhatdan:
I think the best you can do is volume mount in both and search them both.

rhatdan:
@giuseppe LGTM

ashcrow:
I'm going to double check that we have a ``daemon.json`` patch for our flags as @rhatdan noted. Please wait on merging until I've verified.

ashcrow:
Looks like both flags may be either CLI flags or configuration file options:
```
530c7df2 daemon/config.go (Dan Walsh               2016-01-22 07:08:35 -0800  99)       BlockedRegistries    []string            `json:"block-registry,omitempty"`
530c7df2 daemon/config.go (Dan Walsh               2016-01-22 07:08:35 -0800 100)       AdditionalRegistries []string            `json:"add-registry,omitempty"`
<snip>
530c7df2 daemon/config.go (Dan Walsh               2016-01-22 07:08:35 -0800 186)       cmd.Var(opts.NewListOptsRef(&config.BlockedRegistries, registry.ValidateIndexName), []string{"-block-registry"}, usageFn("Don't contact given registry"))
530c7df2 daemon/config.go (Dan Walsh               2016-01-22 07:08:35 -0800 187)       cmd.Var(opts.NewListOptsRef(&config.AdditionalRegistries, registry.ValidateIndexName), []string{"-add-registry"}, usageFn("Registry to query before a public one"))
```

giuseppe:
@rh-atomic-bot r+ 998a01c

richm:
If log-driver is defined in both /etc/sysconfig/docker as `--log-driver=something` and log-driver is defined in /etc/docker/daemon.json as `"log_driver": "something"`, which one takes precedence?

ashcrow:
@richm docker will fail to start in that case if using a traditional package install. If it is defined in both locations but runs via a system container then ``daemon.json`` would take precedence as it will ignore ``/etc/sysconfig/docker``.

Because of this ``daemon.json`` existing and having ``log_driver`` would mean it has precedence.

PR #66, opened 2017-05-09
Prepare docker-fedora for fedora registry

yuqi-zhang:
Add proper labels and README for docker-fedora image, intended to
be built into the fedora registry as "container-engine" system
container.

### DIFF
--- a/docker-fedora/README.md
+++ b/docker-fedora/README.md
@@ -1,0 +1,58 @@
+# container-engine
+
+This is the container engine system container for fedora systems based on the moby project.
+For more information refer to: https://github.com/moby/moby
+
+Note this only provides the *daemon process*, and does not include CLI tools. The current use
+case for this would be switching between different versions of the engine by updating the
+container.
+
+## Building the image from source:
+
+```
+# git clone https://github.com/projectatomic/atomic-system-containers
+# cd atomic-system-containers/docker-fedora
+# docker build -t container-engine .
+```
+
+## Running the system container, with the atomic CLI:
+
+Pull from registry into ostree:
+
+```
+# atomic pull --storage ostree $REGISTRY/container-engine
+```
+
+Or alternatively, pull from local docker:
+
+```
+# atomic pull --storage ostree docker:container-engine:latest
+```
+
+Install the container:
+
+Currently we recommend using --system-package=no to avoid having rpmbuild create an rpm file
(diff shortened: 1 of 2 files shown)


yuqi-zhang:
The README probably needs to be fleshed out a bit. @ashcrow PTAL if the labels and naming are what was discussed.

docker-fedora/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+```
+# git clone https://github.com/projectatomic/atomic-system-containers
+# cd atomic-system-containers/docker-fedora
+# docker build -t container-engine .
+```
+
+## Running the system container, with the atomic CLI:
+
+Requirement: etcd and flannel services are running
```
ashcrow: I don't think that these are needed for the image to work. They would be needed for origin/openshift/etc.. though.

yuqi-zhang: Good point, I will remove those.

Speaking of which, does the container require rpmbuild on the host to install?

ashcrow: @yuqi-zhang good question. As a system container it does not. As an rpm'd system container it might. @giuseppe do you know?

yuqi-zhang:
@giuseppe PTAL. I am not super familiar with some of the details so please add anything you see fit to the README

docker-fedora/README.md
```diff
@@ -0,0 +1,54 @@
...
+Or alternatively, pull from local docker:
+
+```
+# atomic pull --storage ostree docker:container-engine:latest
+```
+
+Install the container:
+
+```
+# atomic install --system --name=container-engine ($REGISTRY)/container-engine
```
ashcrow: For the system container work in origin we are using ``--system-package=no``, but this should work as well.

giuseppe: yeah let's enforce that until https://github.com/projectatomic/atomic/issues/981 is sorted out

yuqi-zhang:
Added the flag to both "usage" label and docs for now

giuseppe:
@yuqi-zhang is this still a WIP?  It looks fine to me and I think we can merge it

yuqi-zhang:
Removed WIP, feel free to merge when ready!

giuseppe:
@rh-atomic-bot r+ 83be636

PR #67, opened 2017-05-12
docker containers: Add /var/run for secrets

ashcrow:
Mount /var/run into the container as rw so secrets can be read.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1450286

### DIFF
--- a/docker-centos/config.json.template
+++ b/docker-centos/config.json.template
@@ -311,6 +311,17 @@
 		"mode=755"
 	    ]
 	},
+	{
+	    "type": "bind",
+	    "source": "/var/run",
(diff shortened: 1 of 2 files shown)


giuseppe:
isn't '/var/run` a symlink to `/run` that is already mounted from the host?

ashcrow:
@giuseppe yes. However openshift and kubernetes look in /var/run for specific files it seems (see https://bugzilla.redhat.com/show_bug.cgi?id=1450286)

ashcrow:
You're right, it's a symlink and ``/run/`` and ``/var/run`` are the same.

PR #68, opened 2017-05-16
docker: use systemd-notify

giuseppe:
Since newer runC have this feature, take advantage of it to properly report when the service is ready

### DIFF
--- a/docker-centos/init.sh
+++ b/docker-centos/init.sh
@@ -18,7 +18,7 @@ do
     $i &
 done
 
-/usr/bin/docker-containerd-current \
+NOTIFY_SOCKET=/dev/null /usr/bin/docker-containerd-current \
     --listen unix:///run/containerd.sock \
     --shim /usr/bin/docker-containerd-shim-current \
     --start-timeout 2m &
--- a/docker-centos/init.sh
+++ b/docker-centos/init.sh
@@ -18,6 +18,8 @@ do
     $i &
 done
 
+# Inhibit sd-notify for docker-containerd, we want to get the notification
+# from the docker process
 NOTIFY_SOCKET=/dev/null /usr/bin/docker-containerd-current \
     --listen unix:///run/containerd.sock \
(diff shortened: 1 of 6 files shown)


docker-centos/init.sh
```diff
@@ -18,7 +18,7 @@ do
     $i &
 done
 
-/usr/bin/docker-containerd-current \
+NOTIFY_SOCKET=/dev/null /usr/bin/docker-containerd-current \
```
rhatdan: What does this do?  How does the NOTIFY_SOCKET get back to systemd?

giuseppe: this is to prevent to get the `READY=1` from docker-containerd.  Doing `NOTIFY_SOCKET=/dev/null` inhibits docker-containerd-current to communicate with the runC socket (which in turn communicates with the systemd socket).

ashcrow: @giuseppe it might be helpful to add a comment above the line with that information. A month from now I'd probably ``git blame`` this line and then find the author so I could figure out what was going on 😸

giuseppe:
@ashcrow I've added two fixup patches that comment the `NOTIFY_SOCKET=/dev/null`part

ashcrow:
@yuqi-zhang did you end up finding out what caused the error?

yuqi-zhang:
@ashcrow unfortunately not, I ran into some weird selinux thing as well but @giuseppe could not replicate, so I will try with a fresh machine today and report back.

ashcrow:
@yuqi-zhang sounds good.

giuseppe:
@rh-atomic-bot delegate=yuqi-zhang

giuseppe:
@yuqi-zhang I added a patch to make /run/systemd rslave and not get the runC notify socket into the host.  Does it solve the problem for you?

yuqi-zhang:
Did some tests, looks good. Thanks @giuseppe !

@rh-atomic-bot r+ aa6851b

PR #72, opened 2017-06-06
docker-fedora: Drop default /etc/oci-umount.conf

ashcrow:
This change drops an empty ```/etc/oci-umount.conf``` if the host doesn't
exist on the host. This is to avoid ```oci-umount``` raising an error and
keeping containers from running.

Ref: https://github.com/rhatdan/oci-umount/issues/2


giuseppe:
LGTM, @yuqi-zhang if it looks fine for you as well, please merge it

yuqi-zhang:
@rh-atomic-bot r+ 075b01b

PR #73, opened 2017-06-07
docker-fedora: /etc/oci-umount.conf is now required

ashcrow:
This file is now required for docker to start. The contents have been
updated to reflect upstream.

Ref: https://github.com/projectatomic/oci-umount/blob/master/oci-umount.conf


giuseppe:
could we copy this file from the container itself?  We will not risk to carry around an outdated version, something like:
`RUN cp /etc/oci-umount.conf /exports/hostfs/etc`

What do you think?

ashcrow:
@giuseppe Oh I like that. Will update.

ashcrow:
@giuseppe PTAL

giuseppe:
LGTM

@rh-atomic-bot r+ eefd8af

PR #75, opened 2017-06-08
Clarify docker and container-engine

yuqi-zhang:
As discussed, the docker system container should still be called
docker to match the package. The container-engine system container,
which was designed as an abstract engine, will be split out. Taking
advantage of the fedora layered build system, it will currently
just layer a README on top of the docker image (since we only support
docker at the moment). In the future, if container-engine gets
expanded, it will be easy to make changes just to the container-engine
container.

### DIFF
--- a/docker-fedora/Dockerfile
+++ b/docker-fedora/Dockerfile
@@ -1,13 +1,13 @@
 FROM registry.fedoraproject.org/fedora:rawhide
 
 ENV VERSION=0 RELEASE=1 ARCH=x86_64
-LABEL com.redhat.component="container-engine" \
-      name="$FGC/container-engine" \
+LABEL com.redhat.component="docker" \
+      name="docker" \
       version="$VERSION" \
       release="$RELEASE.$DISTTAG" \
       architecture="$ARCH" \
-      usage="atomic install --system --system-package=no container-engine && systemctl start container-engine" \
-      summary="The docker container engine as a system container." \
+      usage="atomic install --system --system-package=no docker && systemctl start docker" \
+      summary="The docker daemon as a system container." \
       maintainer="Giuseppe Scrivano <gscrivan@redhat.com>" \
       atomic.type="system"
 
--- a/docker-fedora/README.md
+++ b/docker-fedora/README.md
@@ -1,18 +1,17 @@
-# container-engine
+# docker
 
-This is the container engine system container for fedora systems based on the moby project.
+This is the docker daemon system container for Fedora systems based on the moby project.
 For more information refer to: https://github.com/moby/moby
 
 Note this only provides the *daemon process*, and does not include CLI tools. The current use
-case for this would be switching between different versions of the engine by updating the
-container.
+case for this would be switching between different versions of docker by updating the container.
 
 ## Building the image from source:
 
 ```
 # git clone https://github.com/projectatomic/atomic-system-containers
 # cd atomic-system-containers/docker-fedora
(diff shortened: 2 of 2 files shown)


yuqi-zhang:
Note that currently you cannot build container-engine since the docker image does not exist yet. To test functionality, you can change the FROM line to something like docker.io/gscrivano/docker-fedora.

Also using f25 as the placeholder there since there doesn't seem to be a registry.fedoraproject.org/rawhide/XXX

container-engine-fedora/README.md
```diff
@@ -0,0 +1,59 @@
+# container-engine
+
+This is the container engine system container for fedora systems. This is meant to be an
```
giuseppe: should "fedora" be uppercase?

ashcrow:
@giuseppe I am not sure either. If it goes to Fedora distgit only I think it needs to be clear this is a system container less someone thinks they can pull it in the traditional method thinking it will provide a service file/configuration.

yuqi-zhang:
I guess it makes more sense to just have it in Fedora distgit. The README and Dockerfile description should hopefully clarify enough that its a system container.

container-engine-fedora/README.md
ashcrow: Probably. That's a hold over from the original readme. If you update this @yuqi-zhang please update the other readme as well.

yuqi-zhang:
Removed the container-engine folder for now.

Now I think about it, I'm somewhat uncertain about the naming scheme of just "docker". Perhaps it should be called "docker-daemon" instead? Since 
1. it's a system container, there may be other containers in the future that wish to use this namespace and 
2. it only provides the daemon, not the whole docker package as the container name might imply

ashcrow:
@yuqi-zhang I think that ``docker-daemon`` would make sense. Tagging in @dustymabe for his thoughts as well.

dustymabe:
> it only provides the daemon, not the whole docker package as the container name might imply

the docker rpm isn't installed in the container? 

> @yuqi-zhang I think that docker-daemon would make sense. Tagging in @dustymabe for his thoughts as well.

i personally would like to keep container names == rpm names where it makes sense. I almost would even like to have an rpm named `container-engine` that just requires docker :). I could be swayed to `docker-daemon`, but would like for there to be good reasons to vary from the name of the rpm.

yuqi-zhang:
docker is installed in the container, but its more so because the container doesn't provide the CLI to the host, just the daemon. I suppose that's fine since you can just do something like `runc exec docker docker ${COMMAND}` through the container.

Now I think about it maybe just keeping it as docker makes sense. What do you think @giuseppe

giuseppe:
we will eventually provide the client as well (copy it to the host as part of the container installation) or at least a wrapper around runc. I think both names are fine, the advantage of something else than `docker` is that it won't conflict with an existing installation of `docker` on the host.

yuqi-zhang:
I think if we choose to eventually copy the client as well, we should just name it docker for now to match the package, so in the future we don't have to change the namespace in the fedora registry. The README should hopefully be enough to alert people of the caveats.

Any objections to just naming it docker @ashcrow @giuseppe ? If not I will open another request in bugzilla.

ashcrow:
No objections here.

yuqi-zhang:
@rh-atomic-bot delegate=giuseppe

giuseppe:
@rh-atomic-bot r+ a95a21f

PR #76, opened 2017-06-14
tests: verify JSON files are valid

giuseppe:

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -6,6 +6,7 @@ packages:
     - findutils
     - git
     - golang-bin
+    - yajl
 
 tests:
     - ./test.sh
--- a/test.sh
+++ b/test.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 set -euo pipefail
 
+find . -name 'config.json*' -o -name 'manifest.json' | while read i; do
+    echo Testing file $i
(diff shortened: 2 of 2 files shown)


yuqi-zhang:
What is this trying to catch specifically? Let's say if I'm missing a comma somewhere in the json file, the lint file will return something like: `ValueError: Expecting , delimiter: line 22 column 4 (char 393)` and this never gets run. If I move this above the lint tests and there is an error, it just boots me out of root.

giuseppe:
@yuqi-zhang good point.  I've moved it above and dropped the quiet flag so we will get a better and more helpful error message when it fails.  What do you think?

yuqi-zhang:
I feel that the error message isn't very descriptive:

> JSON is valid
JSON is valid
JSON is valid
JSON is valid
parse error: after array element, I expect ',' or ']'
                                        { 	"ociVersion": "1.0.0", 	"pl
                     (right here) ------^
JSON is invalid

Also boots me out of root still, which is a minor annoyance.
Sorry for the nits.

giuseppe:
thanks for the review, I've added the file name to the output now

yuqi-zhang:
I think this is better than the old error that you would see.

@rh-atomic-bot r+ 415caf0

PR #78, opened 2017-06-20
etcd tweaks for running with kube ansible

jasonbrooks:
* add etcdctl to host
* add cap for reading cert dir
* mount config/certs dir
* copy and source config file from rpm rather than export all the values

### DIFF
--- a/etcd/Dockerfile
+++ b/etcd/Dockerfile
@@ -22,6 +22,13 @@ COPY uninstall.sh /usr/bin/uninstall.sh
 
 COPY tmpfiles.template config.json.template service.template manifest.json /exports/
 
+RUN mkdir -p /exports/hostfs/usr/local/bin/ && cp /usr/bin/etcdctl /exports/hostfs/usr/local/bin/etcdctl
+
+# RUN mkdir -p /exports/hostfs/usr/local/bin/ && echo "runc exec -- etcd /usr/bin/etcdctl \$@" \
+# > /exports/hostfs/usr/local/bin/etcdctl && chmod +x /exports/hostfs/usr/local/bin/etcdctl
+
+RUN mkdir -p /exports/hostfs/etc/etcd && cp /etc/etcd/etcd.conf /exports/hostfs/etc/etcd
+
 EXPOSE 4001 7001 2379 2380
 
 CMD ["/usr/bin/etcd-env.sh", "/usr/bin/etcd"]
--- a/etcd/Dockerfile
+++ b/etcd/Dockerfile
@@ -24,10 +24,7 @@ COPY tmpfiles.template config.json.template service.template manifest.json /expo
 
 RUN mkdir -p /exports/hostfs/usr/local/bin/ && cp /usr/bin/etcdctl /exports/hostfs/usr/local/bin/etcdctl
 
-# RUN mkdir -p /exports/hostfs/usr/local/bin/ && echo "runc exec -- etcd /usr/bin/etcdctl \$@" \
-# > /exports/hostfs/usr/local/bin/etcdctl && chmod +x /exports/hostfs/usr/local/bin/etcdctl
-
-RUN mkdir -p /exports/hostfs/etc/etcd && cp /etc/etcd/etcd.conf /exports/hostfs/etc/etcd
+RUN mkdir -p /exports/hostfs/etc/etcd && sed -e "/^ETCD_DATA_DIR/d" -e s"|^ETCD_NAME=|#ETCD_NAME=|" < /etc/etcd/etcd.conf > /exports/hostfs/etc/etcd/etcd.conf
 
 EXPOSE 4001 7001 2379 2380
 
--- a/etcd/config.json.template
+++ b/etcd/config.json.template
@@ -59,26 +59,31 @@
 		"cwd": "/",
 		"capabilities": {
 			"bounding": [
+                                "CAP_DAC_READ_SEARCH",
 				"CAP_AUDIT_WRITE",
 				"CAP_KILL",
 				"CAP_NET_BIND_SERVICE"
 			],
 			"permitted": [
+                                "CAP_DAC_READ_SEARCH",
                                 "CAP_AUDIT_WRITE",
                                 "CAP_KILL",
                                 "CAP_NET_BIND_SERVICE"
                         ],
 			"inheritable": [
+                                "CAP_DAC_READ_SEARCH",
                                 "CAP_AUDIT_WRITE",
                                 "CAP_KILL",
                                 "CAP_NET_BIND_SERVICE"
                         ],
 			"effective": [
+                                "CAP_DAC_READ_SEARCH",
                                 "CAP_AUDIT_WRITE",
                                 "CAP_KILL",
                                 "CAP_NET_BIND_SERVICE"
                         ],
 			"ambient": [
+                                "CAP_DAC_READ_SEARCH",
                                 "CAP_AUDIT_WRITE",
                                 "CAP_KILL",
                                 "CAP_NET_BIND_SERVICE"
@@ -199,7 +204,17 @@
                                 "rbind",
                                 "rprivate"
                         ]
-        }
+                 },
+                 {
+                         "type": "bind",
+                         "source": "/etc/etcd/",
+                         "destination": "/etc/etcd/",
+                         "options": [
+                                 "rbind",
+                                 "ro",
+                                 "rprivate"
+                         ]
+                 }
 	],
 	"linux": {
 		"resources": {
--- a/etcd/config.json.template
+++ b/etcd/config.json.template
@@ -17,6 +17,7 @@
 		"env": [
 			"NAME=$NAME",
 			"ETCD_NAME=$ETCD_NAME",
+                        "ETCD_DATA_DIR=$ETCD_DATA_DIR",
 			"ETCD_ADVERTISE_CLIENT_URLS=$ETCD_ADVERTISE_CLIENT_URLS",
 			"ETCD_LISTEN_CLIENT_URLS=$ETCD_LISTEN_CLIENT_URLS",
 			"ETCD_INITIAL_ADVERTISE_PEER_URLS=$ETCD_INITIAL_ADVERTISE_PEER_URLS",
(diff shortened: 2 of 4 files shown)


etcd/etcd-env.sh
```diff
@@ -6,11 +6,11 @@ if test x$NAME == x; then
...
 export ETCD_NAME=${ETCD_NAME:-$HOSTNAME}
 export ETCD_DATA_DIR=/var/lib/etcd/${NAME}.etcd
-export ETCD_ADVERTISE_CLIENT_URLS=${ETCD_ADVERTISE_CLIENT_URLS:-http://${ipaddress}:2379,http://${ipaddress}:4001}
-export ETCD_LISTEN_CLIENT_URLS=${ETCD_LISTEN_CLIENT_URLS:-http://0.0.0.0:2379,http://0.0.0.0:4001}
-export ETCD_INITIAL_ADVERTISE_PEER_URLS=${ETCD_INITIAL_ADVERTISE_PEER_URLS:-http://${ipaddress}:2380,http://${ipaddress}:7001}
-export ETCD_LISTEN_PEER_URLS=${ETCD_LISTEN_PEER_URLS:-http://0.0.0.0:2380,http://0.0.0.0:7001}
-export ETCD_INITIAL_CLUSTER=${ETCD_INITIAL_CLUSTER:-$HOSTNAME=http://${ipaddress}:2380,$HOSTNAME=http://${ipaddress}:7001}
+
+sed -i s/ETCD_NAME.*/ETCD_NAME=$ETCD_NAME/g /etc/etcd/etcd.conf
+sed -i s/ETCD_DATA_DIR.*/ETCD_DATA_DIR=$ETCD_DATA_DIR/g /etc/etcd/etcd.conf
```
strigazi: Why not get these from the configuration?

etcd/etcd-env.sh
```diff
@@ -6,11 +6,11 @@ if test x$NAME == x; then
 fi
 export ETCD_NAME=${ETCD_NAME:-$HOSTNAME}
 export ETCD_DATA_DIR=/var/lib/etcd/${NAME}.etcd
```
strigazi: I think the ETCD_DATA_DIR needs to be configurable.

The use case we have is that we want to store the etcd data in an external block device mounted to the host for persistency. Even if we don't want to store etcd's data in an external device we may need to backup etcd.

Thoughts?

strigazi: Maybe also add CAP for /var/lib/etcd ?

jasonbrooks: This is how I got it to work -- there's a relationship between what's going on in config.json.template, etcd-env.sh, and tmpfiles.template where the data dir's name is tied to the name we give to the container when we install it. I've tried a couple of variations without success. Can you shed additional light on this, @giuseppe

ashcrow: You could create a variable in ``manifest.json`` called ``DATA_DIR`` or something similar. ``DATA_DIR`` would then be used in the templates and shell script. The default could be ``/var/lib/$NAME``. If ``DATA_DIR`` is overridden then it would point to the provided location instead.

jasonbrooks: Ah, ok, I tried something like this but didn't mod manifest.json -- I'll try w/ that

jasonbrooks: OK, now it uses the data dir value that's in etcd.conf, which is set in the package to be `/var/lib/etcd/default.etcd`. You can change that to another value, as long as it's under `/var/lib/etcd`. I tried to set the default in `manifest.json` to be based on $NAME, but either it wasn't getting populated correctly, or it was getting overwritten by the default in the config file.

etcd/etcd-env.sh
giuseppe: ETCD_DATA_DIR is always at the same location in the container (`/var/lib/etcd`), it wouldn't make sense to allow its configuration

etcd/etcd-env.sh
giuseppe: we already have `STATE_DIRECTORY` for that.  It gets a default value from `atomic` which is `/var/lib` when installing a system container and `%HOME/.data` when using `--user`.  Even if it gets a value from atomic, ${STATE_DIRECTORY} can be overriden at installation time, so you can do:

```bash
atomic install --system --set STATE_DIRECTORY=/mnt/somewhere/else etcd
```

and the etcd container will use the storage from `/mnt/somewhere/else/${NAME}.etcd`

It will be nice to allow to set the complete storage directory (without the ${NAME}.etcd part) but for allowing it and at the same time don't lose the capability of running multiple etcd instances, I think we will need to modify atomic to support variables expansion in the `defaultValues` block as well.

etcd/config.json.template
```diff
@@ -199,7 +204,17 @@
                                 "rbind",
                                 "rprivate"
                         ]
-        }
+                 },
+                 {
+                         "type": "bind",
+                         "source": "/etc/etcd/",
```
giuseppe: `/etc/$NAME`for "source" given my other comment for manifest.json

giuseppe:
sorry I had some comments already yesterday.  I've just forgot to send the review and they were waiting in the github limbo.

etcd/Dockerfile
```diff
@@ -22,6 +22,13 @@ COPY uninstall.sh /usr/bin/uninstall.sh
 
 COPY tmpfiles.template config.json.template service.template manifest.json /exports/
 
+RUN mkdir -p /exports/hostfs/usr/local/bin/ && cp /usr/bin/etcdctl /exports/hostfs/usr/local/bin/etcdctl
+
+# RUN mkdir -p /exports/hostfs/usr/local/bin/ && echo "runc exec -- etcd /usr/bin/etcdctl \$@" \
+# > /exports/hostfs/usr/local/bin/etcdctl && chmod +x /exports/hostfs/usr/local/bin/etcdctl
+
+RUN mkdir -p /exports/hostfs/etc/etcd && cp /etc/etcd/etcd.conf /exports/hostfs/etc/etcd
```
jasonbrooks: I tried this, and this works, but what about the data dir? By default, each installation of the container will use the same host location for the data dir, unless the user changes it in the config file. The user would also have to change the ports in the config file to run two of these together, so maybe this isn't a problem?

ashcrow:
Latest changes look good @jasonbrooks!

etcd/config.json.template
```diff
@@ -177,7 +178,7 @@
 		},
 		{
 			"type": "bind",
-			"source": "${STATE_DIRECTORY}/etcd/${NAME}.etcd",
+			"source": "${STATE_DIRECTORY}/etcd",
```
giuseppe: I'd drop this change, so that each instance has its own directory

etcd/etcd-env.sh
```diff
@@ -1,15 +1,5 @@
...
-if test x$NAME == x; then
-    NAME=$HOSTNAME
-fi
-export ETCD_NAME=${ETCD_NAME:-$HOSTNAME}
-export ETCD_DATA_DIR=/var/lib/etcd/${NAME}.etcd
-
-sed -i s/ETCD_NAME.*/ETCD_NAME=$ETCD_NAME/g /etc/etcd/etcd.conf
-sed -i s/ETCD_DATA_DIR.*/ETCD_DATA_DIR=$ETCD_DATA_DIR/g /etc/etcd/etcd.conf
-
 source /etc/etcd/etcd.conf
```
giuseppe: same here, I'd not remove these lines

giuseppe:
I think we should drop the "patch 2" if you have nothing against.  After these changes will `--set` still work when specified at installation time?  I can verify this tomorrow morning my time

giuseppe:
it seems to work well!

I'd suggest only these tiny changes on top of yours:

```diff
diff --git a/etcd/Dockerfile b/etcd/Dockerfile
index d93f4f0..004a94b 100644
--- a/etcd/Dockerfile
+++ b/etcd/Dockerfile
@@ -24,10 +24,7 @@ COPY tmpfiles.template config.json.template service.template manifest.json /expo
 
 RUN mkdir -p /exports/hostfs/usr/local/bin/ && cp /usr/bin/etcdctl /exports/hostfs/usr/local/bin/etcdctl
 
-# RUN mkdir -p /exports/hostfs/usr/local/bin/ && echo "runc exec -- etcd /usr/bin/etcdctl \$@" \
-# > /exports/hostfs/usr/local/bin/etcdctl && chmod +x /exports/hostfs/usr/local/bin/etcdctl
-
-RUN mkdir -p /exports/hostfs/etc/etcd && cp /etc/etcd/etcd.conf /exports/hostfs/etc/etcd
+RUN mkdir -p /exports/hostfs/etc/etcd && sed -e "/^ETCD_DATA_DIR/d" -e s"|^ETCD_NAME=|#ETCD_NAME=|" < /etc/etcd/etcd.conf > /exports/hostfs/etc/etcd/etcd.conf
 
 EXPOSE 4001 7001 2379 2380
 
diff --git a/etcd/manifest.json b/etcd/manifest.json
index 5679ca5..a8bc0d0 100644
--- a/etcd/manifest.json
+++ b/etcd/manifest.json
@@ -42,7 +42,8 @@
        "ETCD_ENABLE_PPROF" : "false"
     },
     "renameFiles" : {
-        "/etc/etcd/etcd.conf" : "$CONF_DIRECTORY/$NAME/etcd.conf"
+        "/etc/etcd/etcd.conf" : "$CONF_DIRECTORY/$NAME/etcd.conf",
+        "/usr/local/bin/etcdctl" : "/usr/local/bin/etcdctl-$NAME"
     }
 }
 
```

1) etcdctl is renamed to the name of the container so multiple instances do not interfere with each other.  
2) drop `ETCD_DATA_DIR` from the configuration file since it makes no sense in the container
3) comment `ETCD_NAME` so that `--set ETCD_NAME=something` still works (but it is possible to override it from the conf file)
4) drop some dead lines

After this change, could you squash the last 3 patches into one?  The last two patches revert the second patch of this series so it will be cleaner to have just the changes from the current version.

etcd/etcd-env.sh
strigazi: I'll give it a go passing STATE_DIRECTORY too.

jasonbrooks:
@giuseppe These changes seem fine, with the exception of renaming etcdctl -- I put it in there for compatibility w/ scripts expecting to be able to run etcdctl, and giving it a different name will break that. Could we do something like if $NAME = etcd don't do the rename? Or rename to $NAMEctl -- in the ansible I use, I call the container etcd so it'll be a drop in replacement.

giuseppe:
unfortunately we can't do the if $NAME part.  OK let's go with ${NAME}ctl, at least it won't block multiple instances of etcd for being installed

PR #80, opened 2017-07-03
SELinuxContext for docker container unit files

ashcrow:
Add ``SELinuxContext=system_u:system_r:container_runtime_t:s0`` to both CentOS and Fedora unit files.


giuseppe:
@rh-atomic-bot r+ e545df9

PR #84, opened 2017-07-11
etcd: check if etcd.conf exists before reading it

giuseppe:
user containers do not copy files to the host.  Check if the file
exists before attempting to read it.

### DIFF
--- a/etcd/etcd-env.sh
+++ b/etcd/etcd-env.sh
@@ -7,7 +7,9 @@ fi
 export ETCD_NAME=${ETCD_NAME:-$HOSTNAME}
 export ETCD_DATA_DIR=/var/lib/etcd/${NAME}.etcd
 
-source /etc/etcd/etcd.conf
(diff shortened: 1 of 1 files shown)


giuseppe:

A simple fix for the `atomic install --user` case, requires these patches in atomic:

https://github.com/projectatomic/atomic/pull/1043

yuqi-zhang:
LGTM, just curious, is this the only relevant case for user containers?

giuseppe:
With the system containers we have now probably yes, as I don't see how we could use other containers as a non privileged user. We could still make it easier for people to have images that work without many problems in both ways but for now this small change allows to run etcd using bubblewrap and bwrap-oci using more or less the same CLI as for the system container case

ashcrow:
I'll let @jasonbrooks do the honors since he has been delegated.

jasonbrooks:
Ah, do I have to do this:

@rh-atomic-bot r+ 0945867

PR #85, opened 2017-07-11
Copy oci-umount.conf if it exists

ashcrow:
- Copy oci-umount.conf to the host if it exists in the container
- Port changes from Fedora git to docker-fedora

### DIFF
--- a/docker-fedora/Dockerfile
+++ b/docker-fedora/Dockerfile
@@ -13,9 +13,8 @@ LABEL com.redhat.component="docker" \
 
 RUN dnf install --setopt=tsflags=nodocs -y docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz oci-register-machine \
     && rpm -V docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz oci-register-machine \
-    && mkdir -p /usr/lib/modules \
-    && dnf clean all \
-    && mkdir -p /exports/hostfs/etc/docker
+    && mkdir -p /usr/lib/modules /exports/hostfs/etc/docker \
+    && dnf clean all
 
 RUN ln -s /usr/libexec/docker/docker-runc-current /usr/bin/docker-runc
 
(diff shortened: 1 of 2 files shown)


ashcrow:
@yuqi-zhang note that a fedora rebuild is not needed with these changes.

docker-fedora/Dockerfile
```diff
@@ -26,5 +28,8 @@ COPY config.json.template service.template tmpfiles.template /exports/
 COPY daemon.json /exports/hostfs/etc/docker/container-daemon.json
 # https://github.com/rhatdan/oci-umount/issues/2
 RUN cp /etc/oci-umount.conf /exports/hostfs/etc
+# Copy config if available
+RUN (test -e /etc/oci-umount.conf && cp /etc/oci-umount.conf /exports/hostfs/etc) || true
```
yuqi-zhang: Shouldn't the line above this be removed?

Also this change is already in f25, but it *shouldn't* be an issue for later versions (as the file will be present)

docker-fedora/Dockerfile
```diff
@@ -13,7 +13,9 @@ LABEL com.redhat.component="docker" \
 
 RUN dnf install --setopt=tsflags=nodocs -y docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz oci-register-machine \
     && rpm -V docker container-storage-setup container-selinux cloud-utils-growpart python-docker-py docker-novolume-plugin lvm2 iptables procps-ng xz oci-register-machine \
-    && mkdir -p /usr/lib/modules && dnf clean all
+    && mkdir -p /usr/lib/modules \
+    && dnf clean all \
+    && mkdir -p /exports/hostfs/etc/docker
```
yuqi-zhang: Formatting feels kind of weird, maybe condense with "/usr/lib/modules" into one line?

yuqi-zhang:
Oh wow that's a scary red X, sorry...

ashcrow:
@yuqi-zhang ⬆️

yuqi-zhang:
LGTM

@rh-atomic-bot delegate=giuseppe

giuseppe:
I thought about having here only the last version (rawhide) and keep differences for different versions in the Fedora dist git.  What do you think about it? Since anyway these changes are not going to break on newer systems, it is fine to merge them as well.. LGTM

giuseppe:
@rh-atomic-bot r+ 0b3863f

yuqi-zhang:
@giuseppe I think that is the plan

PR #87, opened 2017-07-13
Create cri-o system container

yuqi-zhang:
Create a system container for cri-o, built from source inside the
container. Currently this is still experimental, but has been
tested to work on f25/26. In the future the build script will
likely be replaced by the packaged version.

The container image comes with crioctl installed on the host, but
does not include cni configs. The daemon will run, but to create
pods/containers, the user must still set up cni configs on the host.

### DIFF
--- a/cri-o/Dockerfile
+++ b/cri-o/Dockerfile
@@ -1,0 +1,24 @@
+FROM registry.fedoraproject.org/fedora:25
+
+ENV VERSION=0 RELEASE=1 ARCH=x86_64
+LABEL com.redhat.component="cri-o" \
+      name="$FGC/cri-o" \
+      version="$VERSION" \
+      release="$RELEASE.$DISTTAG" \
+      architecture="$ARCH" \
+      usage="atomic install --system --system-package=no crio && systemctl start crio" \
+      summary="The cri-o daemon as a system container." \
+      maintainer="Yu Qi Zhang <jzehrarnyg@gmail.com>" \
+      atomic.type="system"
+
+RUN dnf -y --setopt=tsflags=nodocs install btrfs-progs-devel device-mapper-devel glib2-devel glibc-devel glibc-static gpgme-devel libassuan-devel libgpg-error-devel libseccomp-devel libselinux-devel \
+    pkgconfig runc wget git findutils gcc iptables skopeo && \
+    dnf clean all
+
+COPY tmpfiles.template config.json.template service.template /exports/
+
+COPY build.sh run.sh /usr/bin/
+
+RUN /usr/bin/build.sh
+
+CMD ["/usr/bin/run.sh"]
--- a/cri-o/Dockerfile
+++ b/cri-o/Dockerfile
@@ -11,10 +11,6 @@ LABEL com.redhat.component="cri-o" \
       maintainer="Yu Qi Zhang <jzehrarnyg@gmail.com>" \
       atomic.type="system"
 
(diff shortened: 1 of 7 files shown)


yuqi-zhang:
A note that the container image is very large (1.4G) at this time. If you want to just test it, I have built it to docker.io/jerzhang/crio

crio/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+      name="$FGC/cri-o" \
+      version="$VERSION" \
+      release="$RELEASE.$DISTTAG" \
+      architecture="$ARCH" \
+      usage="atomic install --system --system-package=no crio && systemctl start crio" \
+      summary="The cri-o daemon as a system container." \
+      maintainer="Yu Qi Zhang <jzehrarnyg@gmail.com>" \
+      atomic.type="system"
+
+RUN dnf -y --setopt=tsflags=nodocs install btrfs-progs-devel device-mapper-devel glib2-devel glibc-devel glibc-static gpgme-devel libassuan-devel libgpg-error-devel libseccomp-devel libselinux-devel \
```
giuseppe: can we move this into build.sh?

We can also uninstall the packages we don't need at the end of build.sh.  It should help to drop the size of the image

giuseppe:
tagging also @rhatdan @mrunalp @runcom @ashcrow

crio/Dockerfile
yuqi-zhang: Yep, doing right now.

One main reason I wasn't very focused on this was that we were going to eventually use the packaged version, which will drop a lot of these packages. There are definitely more optimizable parts once we switch to that.

rhatdan:
Why not use the crio package that we have built for f25?
Replacing the Dockerfile with a buildah script would be kind of cool...

yuqi-zhang:
I built this initially for testing, since the package is still in testing in bodhi: https://bodhi.fedoraproject.org/updates/?packages=cri-o

yuqi-zhang:
@giuseppe I removed the go binaries as well, now down to 500M

ashcrow:
I agree with @rhatdan. It would be best to use the rpm in the system container once it's available, but the temporary build makes sense for testing.

@runcom is https://bodhi.fedoraproject.org/updates/FEDORA-2017-48626b6be8 ready to be pushed to stable?

runcom:
Pushed

rhatdan:
@ashcrow I just had @lsm5 push it to stable.

giuseppe:
what about something like?

RUN dnf install -y cri-o || /usr/bin/build.sh

for the time being so we can use this on Fedora 25 as well

ashcrow:
@giuseppe the F25 and F26 packages have been tagged stable by @runcom

yuqi-zhang:
I'm pretty sure tagged stable != pushed to stable. I can update the Dockerfile/build.sh once the cri-o rpm lands.

ashcrow:
@yuqi-zhang you're correct. Tagging doesn't make it stable but it should move into stable repo in the next iteration or so.

giuseppe:
@yuqi-zhang if it is going to be landed soon, let's just use the rpm package, it is not a big deal if it doesn't build on F25 yet

giuseppe:
we decided with @yuqi-zhang that I'd keep working on this cleanup for a first version, I opened a PR here that adds some patches:

https://github.com/projectatomic/atomic-system-containers/pull/88

ashcrow:
@giuseppe should we close this in favor of work going on in #88?

giuseppe:
Yes we can, all the changes are made there

PR #89, opened 2017-07-17
net-snmp: new container

giuseppe:
still working on it, a first implementation for https://trello.com/c/0JnUTpoJ/784-net-snmp-system-container


net-snmp/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+ADD run.sh /usr/bin/
+
+# System container files
+COPY tmpfiles.template service.template manifest.json config.json.template /exports/
+
+RUN mkdir -p /exports/hostfs/etc/{sysconfig,snmp} /exports/hostfs/var/lib/net-snmp/ && \
+    cp /etc/sysconfig/snmpd /etc/sysconfig/snmptrapd /exports/hostfs/etc/sysconfig/ && \
+    cp /etc/snmp/snmpd.conf /etc/snmp/snmptrapd.conf /exports/hostfs/etc/snmp/ && \
+    echo "# WAIT_FOR_CONFIGURATION=0" >> /exports/hostfs/etc/sysconfig/snmpd
```
ashcrow: Wait for it

net-snmp/run.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+(
+    . /etc/sysconfig/snmpd
+    if test "$WAIT_FOR_CONFIGURATION" == 1; then
+        systemd-notify --ready
+        exec sleep 9999d
```
ashcrow: Mind commenting this line?

giuseppe: sure.  I'd like to drop this hack though.  I didn't manage to find a way to create an user using the container.  The issue is that I cannot use "net-snmp-create-v3-user" when the daemon is running (it is not possible it seems), so I cannot "runc exec " into the container to create an user since there is no container running.  This hack is to set an environment variable so that the container just hangs and it is possible to "runc exec" into it and run commands.

We should probably find a way to run oneshot containers here

ashcrow: So are you currently making the user manually outside of the container?

giuseppe: yes, it must be created interactively since you need to specify user/password (and can create multiple users).  The configuration should usually be between the `atomic install --system` and the `systemctl start $NAME`, just in this case it is more difficult since it requires tools that are inside the container image and at the same time we cannot run the daemon.

ashcrow: @giuseppe what about having ``run.sh`` check to see if the user is created. If it doesn't exist, it creates and runs the daemons. If it does exist, it goes directly to starting the daemons? The creation of the user can be looked at with something like ```egrep 'r[wo]?user' /etc/snmp/snmpd.conf EXPECTED_USER```.

giuseppe: well that is just one example of the settings required, for example the system could require more users or any kind of setting.  I think the best way to handle this is to have some sort of thrown away runonce container here which is how I'd expect it to be handled in the Docker case as well: `docker run --rm -v /etc... -v /var/lib/net-snmp net-snmp-create-v3-user -X PASSWORD EXPECTED_USER`

net-snmp/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM registry.fedoraproject.org/fedora:26
+
+ENV VERSION=0.1 RELEASE=8 ARCH=x86_64
```
TomasTomecek: labels seem to be missing

giuseppe: thanks, I've pushed a fixup patch to add labels

net-snmp/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM registry.fedoraproject.org/fedora:26
+
+
+LABEL maintainer="Giuseppe Scrivano <gscrivan@redhat.com>" \
+      name="net-snmp" \
+      version="0.1" \
+      atomic.type="system" \
+      architecture="x86_64"
+
+ENV VERSION=0.1 RELEASE=8 ARCH=x86_64
```
TomasTomecek: these are still not utilized in labels, so there is no point in defining them

giuseppe:
@TomasTomecek thanks, I've dropped the ENV directive

net-snmp/run.sh
ashcrow: @TomasTomecek provided an interesting idea for a hook in his PR.

ashcrow:
In the mean time I'll put together an admin command system container to go along as well.

giuseppe:
I thought of solving this case by adding an `atomic run $container $command` for system containers (for the time being we could just not support it for the `--user` case).

If the container is already running we would simply run `runc exec $container $command`, while if it is not running we would need to create a temporary `config.json` with a modified `rootfs` and `args`.  With this feature we could do something like:

```
atomic install --system net-snmp
atomic run net-snmp net-snmp-create-v3-user -a password -x encyption-password admin-user'
systemctl start net-snmp
```

What do you think?

ashcrow:
@giuseppe Makes sense. I'm about done with another implementation (closer to the original suggestion). I think may be able to merge the two ideas together and get something snazzy for the user.

ashcrow:
The only thing I'm not keen on is the verbosity using atomic in the command. I feel like administrators would dislike how much extra typing would be involved in what normally would be command and args. However, I think the idea is super sound. We may be able to combine the Atomic feature @giuseppe proposed with a shell script/wrapper/etc that would create simpler end user commands.

ashcrow:
Tagging in @cooktheryan as well as he tends to have a good eye for ops related interfaces.

ashcrow:
I'll take a gander at what it would look like for adding a ``run`` like command to ``atomic`` for running commands inside of system containers.

cooktheryan:
I'm following along with this thread. I love the idea about bring this type of image to atomic. Interested to see how the net-snmp user is handled

ashcrow:
@giuseppe I think we may want to provide another set of labels for running specific commands. Here is a quick idea on it based off your ``run`` example and my secondary image example....

## New Labels
The names can be modified, but provided as examples:

- ``atomic.command.config``: Name of the ``config.json`` that would be used for commands.
- ``atomic.commands.aliases``: Name of commands that the container handles

### Example
This would be a container that would handle the ``net-snmp-create-v3-user`` and ``snmpstatus`` commands:

```
atomic.command.config="config.command.json" \
atomic.command.aliases="net-snmp-create-v3-user,snmpstatus" \
```

## Atomic install addition
``atomic install`` would be extended to create shim for the aliases in ``atomic.command.aliases``. As an example, it would lay down ``net-snmp-create-v3-user`` and ``snmpstatus`` files which really call the proper ``atomic`` command for running the command.

## Option 1: Atomic command
The ``atomic`` command would have a new target called ``command``. This would use the ``atomic.command.config`` label to use the proper config.json file, verify that the command requested is in the ``atomic.command.aliases`` label, and then essentially do ``atomic install`` passing in ``--set COMMAND=$COMMAND `` and ``--ARGS=$ARGS``.

## Option 2: Another install adtion
The ``atomic install`` command would be extended so that the config.json could be switched to use what's in ``atomic.command.config``. The verification of the command would be handled by the shim. ``--set COMMAND=$COMMAND `` and ``--ARGS=$ARGS`` would be used to pass the command and arguments in the container.

## What the user sees when running the system container
```
# net-snmp-create-v3-user -a SHA testing
Enter authentication pass-phrase: 
test
Enter encryption pass-phrase: 
  [press return to reuse the authentication pass-phrase]

adding the following line to /var/lib/net-snmp/snmpd.conf:
   createUser testing SHA "test" DES
adding the following line to /etc/snmp/snmpd.conf:
   rwuser testing
```

giuseppe:
@ashcrow thanks for the detailed explanation!  One problem I see with this approach is that we will require to checkout the system container each time we want to run a command.  Also, we will need to ensure the image used for running the command is the same version that the container is currently running (and the image could have been already deleted from the storage).

I still think it is easier to expand atomic with a new `run` command for system containers.   If `run` is confusing as it is currently used with the `RUN` label for Docker, we could probably add a new command `exec` to atomic.  In the Docker case it will be a wrapper around `docker exec`.

So far we never needed such abstraction as we could use directly `runc exec` but the net-snmp poses a new problem since the commands need to be executed when the daemon is not running.  An `atomic exec` abstraction + handling the case when the container is not running addresses this issue IMO in a more generic way.

To simplify the usage with net-snmp, we could install a wrapper on the host `net-snmp-create-v3-user`, that can be as simple as:

```
#!/bin/sh
atomic exec $NAME net-snmp-create-v3-user $@
```

With this wrapper, users of the system container can use `net-snmp-create-v3-user` in a system container in the same way as it was used before.

TomasTomecek:
Pretty interesting proposals. Did you guys figured out which of the proposals you would like to do?

giuseppe:
I've some WIP for my proposal here: https://github.com/giuseppe/atomic/tree/implement-exec

Now I can do something like:

```
# atomic install --system net-snmp
# mkdir /run/net-snmp  ### This is created by systemd, we will fix it in the tmpfiles file
# atomic exec --storage ostree net-snmp net-snmp-create-v3-user -a password -x encyption-password admin-user
adding the following line to /var/lib/net-snmp/snmpd.conf:
   createUser admin-user MD5 "password" DES encyption-password
adding the following line to /etc/snmp/snmpd.conf:
   rwuser admin-user
```

If we want, we can later simplify the entrypoint to the container with a wrapper script

ashcrow:
@giuseppe @TomasTomecek I'm on board with that idea. I feel strongly in simplifying the command for the system administrator but, as noted, we can add it later.

ashcrow:
@giuseppe I started to put together some unit testing for your branch. I'll probably send a PR your way tomorrow.

TomasTomecek:
yup, looks good! I'll update my PR then

giuseppe:
Opened a PR here: https://github.com/projectatomic/atomic/pull/1075

rhatdan:
I think this could be confusing to users.  atomic run != docker run, the basic idea is to handle an existing container differently.
* Container does not exist
 atomic run == docker run
* Container exists but is not running
atomic run == docker start + docker exec
* Container exists and is running
atomic run == docker exec

It looks to me that you are just doing a version of this with atomic exec, since you say you create a container if it does not exist.

If atomic exec == docker exec, then I could live with it.

What about an option to just extend atomic run to ignore the RUN Label in the case of system containers? Would this satisfy your need without having to add another potentially confusing name?

TomasTomecek:
I actually never understood why the container needs to be running in order to do `docker exec`. This is why I love design of buildah where `buildah run` is actually `run + exec` within a single command. Could be interesting to use the same design in `atomic`.

rhatdan:
The `buildah run`, thing was not by design, I don't believe.  buildah run was supposed to mimic `RUN` syntax in Dockerfile. IE be for a single command, but people have abused/taken advantage of it.

I don't believe if you did two `buildah run` in different terminals, you would end up in the same container, IE YOu would have two containers running.  Where as two runc exec would put you into the same container, as would two atomic runs.

giuseppe:
@rhatdan I thought about extending `atomic run` as the first option (https://github.com/projectatomic/atomic-system-containers/pull/89#issuecomment-323176847), the issue is that `atomic run` expects an IMAGE as its argument, instead here we are using a CONTAINER, and it seemed a bit confusing that the command behaves differently for two backends.

I am fine to move this feature under `run` and add a `--storage` option there, any idea on how we could solve the CONTAINER vs IMAGE difference in the command?

rhatdan:
atomic run --name=CONTAINER IMAGE, correct?
If the IMAGE Name == Container Name, then it just works.
Is  your Container name different then the image name?

giuseppe:
@rhatdan yes, it could be different, so it is safer to assume the container name in any system container command.

Interesting, @baude already added a system container implementation for `run` that assumes NAME is a container name and not an IMAGE, I will take a look at that and see if I can implement the `exec` logic there

rhatdan:
Does atomic command already understand that this is a system container?  Did you add a label to the systme container image, to tell atomic to default the image to a system container?

giuseppe:
yes, this container image already has a `atomic.type="system"` label specified

giuseppe:
I removed the `WAIT_FOR_CONFIGURATION` hack since we are going to fix this use case in `atomic run`

TomasTomecek:
@giuseppe do you think it makes sense to add documentation on how people using this container are meant to create the net-snmp users using the new functionality in `atomic run`?

giuseppe:
@TomasTomecek yes sure, I'll do that as soon as the change is merged into atomic

ashcrow:
Doc commit https://github.com/ashcrow/atomic-system-containers/blob/3b87cd6434cf5d22639b1be8bdceda5cd3b00efd/net-snmp/README.md

giuseppe:
@ashcrow thanks!  I've added a small fixup patch to `s|net-netsnmp|net-snmp|`

ashcrow:
@rh-atomic-bot r+ f45f20f

PR #91, opened 2017-07-19
k8s: Do not sort configuration arguments

MatMaul:
If a parameter is not set using the pattern --param=value but
--param value, the names and values are shuffled.

eg '--cgroup-driver=systemd' VS '--cgroup-driver systemd'


giuseppe:
bot, test pull request

giuseppe:

strigazi:
repro:
https://paste.fedoraproject.org/paste/yvcbMZuKIGM325AcOxJbYg

jasonbrooks:
I initially put in the sorting so that if you supplied arguments on the docker run line (these containers are meant to be runnable as system containers or via docker directly) the arguments from the run line would take precedence over the ones in the config file. 

I suppose, though, that if we're using it in a context where we don't want to look at the config files, we can ignore this run script. In that case I don't think we'd need `$@` among the arguments.

jasonbrooks:
We could change the delimiter from `=` to ` --` and toss in a couple of seds to get around the fact that `sort` doesn't do multicharacter delimiters:

```
ARGS=$(echo "$@ $TEMP_KUBELET_ARGS $KUBE_LOGTOSTDERR $KUBE_LOG_LEVEL $KUBELET_API_SERVER $KUBELET_ADDRESS $KUBELET_PORT $KUBELET_HOSTNAME $KUBE_ALLOW_PRIV $KUBELET_ARGS" | sed 's/ --/|/g' | xargs -n1 | sort -u -t '|' -k 1,1 | xargs | sed 's/|/ --/g')
```

strigazi:
I'm a little confused.

I get that we want the args passed with docker run need to take precedence and "overwrite" the ones in the config file. And it's a very desirable functionality.

To achieve we can put $@ in front (or end, @MatMaul is testing)  of the all the other args.

I don't understand why we need to sort though.

jasonbrooks:
Yeah, this was a couple years ago at this point, I honestly don't remember all the details of why I did it -- I think we're good to take it out. 

@rh-atomic-bot r+ 0ad75c0

MatMaul:
Great thanks for the fast feedback. Here is a small gift, commercial red hats (not from RedHat) everywhere yesterday in a festival at the same moment you were merging my PRs ;)
Unfortunately no more left so I couldn't take some for a nice picture at CERN.
https://photos.app.goo.gl/UFqMCzxHNjuiRfcs2

PR #92, opened 2017-07-20
Mount /dev from host to allow block storage support

MatMaul:
We need access to disk in dev. We also need some basic tools for formatting.
Mounting mqueue is not working I get a busy error that I couldn't solve.
I don't think we care that much since we are sharing the host IPC namespace anyway.


giuseppe:
bot, test pull request

jasonbrooks:
@rh-atomic-bot r+ 8623739

PR #93, opened 2017-07-21
bind mount /run for iptables-restore

jasonbrooks:
Resolves issue found with kube 1.7.1: Failed to execute iptables-restore: failed to open iptables lock /run/xtables.lock: open /run/xtables.lock: read-only file system


giuseppe:
@rh-atomic-bot r+ 5cf413a

PR #94, opened 2017-07-27
ci: rename file and bump to f26

jlebon:
Rename YAML file to the new name, and use the F26 container from the
Fedora registry for tests.


ashcrow:
@rh-atomic-bot r+ 7307730

PR #97, opened 2017-07-31
Docker/CRI-O: ensure /usr/share/rhel exists

giuseppe:
If docker is not installed on the host the `/usr/share/rhel` directory might not be present and runC fails to bind it.  Ensure it exists before the container starts.

### DIFF
--- a/cri-o-centos/tmpfiles.template
+++ b/cri-o-centos/tmpfiles.template
@@ -1,3 +1,4 @@
 d    ${RUN_DIRECTORY}/${NAME}               -        -           -       - -
 d    /etc/crio - - - - -
 Z    /etc/crio - - - - -
+d    /usr/share/rhel - - - - -
--- a/cri-o-fedora/tmpfiles.template
+++ b/cri-o-fedora/tmpfiles.template
@@ -1,3 +1,4 @@
 d    ${RUN_DIRECTORY}/${NAME}               -        -           -       - -
 d    /etc/crio - - - - -
 Z    /etc/crio - - - - -
+d    /usr/share/rhel - - - - -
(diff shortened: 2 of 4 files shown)


cgwalters:
This is assuming that we'll have docker in AH going forward; if we ever removed it, this wouldn't work since we have `/usr` immutable there.

I think there's a deeper problem here; isn't there a more generic version of `/usr/share/rhel` now that's agnostic across distribution/container runtime?  Ideally somewhere in `/etc`?

ashcrow:
@cgwalters agreed that this would be a quick fix that would need to be revisited as ``docker`` and ``crio`` move to a new location and AH drops ``docker`` from the base install.

ashcrow:
Looking at the latest package builds I see links back to ``/etc`` from ```/usr/share/rhel/secrets/``` but currently ```/usr/share/rhel/secrets/``` is still needed AFAICT.

ashcrow:
@rh-atomic-bot r+ 14ed8d3

ashcrow:
@jwhonce Do you know if there is going to be a move to another location other than ```/usr/share/rhel`` for ``secrets`` in the near term?

giuseppe:
I agree that `/usr` doesn't look like the best place for keeping this data.  @rhatdan Is it enough to change this location in the Docker .rpm and container/secrets.go or is there anything more involved?

ashcrow:
@giuseppe After looking deeper: [/etc/containers/secret/](https://github.com/projectatomic/docker/blob/docker-1.12.2/container/secrets.go#L86). I'm looking to see if we can use it.

ashcrow:
@giuseppe I took a look at the current builds and ```/etc/containers/secrets``` do not currently get laid down or populated. However, the code seems to indicate that we could use it without a problem once it does get populated.

rhatdan:
Right /usr/share/rhel/secrets and /etc/containers/secrets should be handled the same way.

PR #99, opened 2017-08-03
docker-fedora: change rootfsPropagation to rslave

giuseppe:
We probably never tried docker-fedora with Origin before but
docker-centos is already using rootfsPropagation and same for CRI-O, so
sync this container with all the other images.


ashcrow:
@rh-atomic-bot r+ 03287cb

PR #100, opened 2017-08-03
docker-centos: drop container-storage-setup as it is not present

giuseppe:


ashcrow:
How was it building with the ``rpm -V`` previously?

giuseppe:
I am afraid it wasn't build for long time.  We should probably check that system containers build, even if it looks quite expensive to build all of them for each PR

ashcrow:
@rh-atomic-bot r+ aeee581

PR #104, opened 2017-08-22
etcd: allow to specify additional bind mounts

giuseppe:

### DIFF
--- a/etcd/config.json.template
+++ b/etcd/config.json.template
@@ -216,6 +216,7 @@
                                  "rprivate"
                          ]
                  }
+                 $ADDTL_MOUNTS
 	],
 	"linux": {
 		"resources": {
--- a/etcd/manifest.json
+++ b/etcd/manifest.json
@@ -39,7 +39,8 @@
 	"ETCD_DEBUG" : "false",
 	"ETCD_LOG_PACKAGE_LEVELS" : "",
 	"ETCD_STRICT_RECONFIG_CHECK" : "false",
-	"ETCD_ENABLE_PPROF" : "false"
+	"ETCD_ENABLE_PPROF" : "false",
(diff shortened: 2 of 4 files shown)


giuseppe:
@ingvagabund this should allow us to replace the dirty hack for etcd in openshift-ansible, as we can specify any bind mount we like

ashcrow:
@giuseppe does this require rebuilds for any current releases or is it forward looking?

giuseppe:
@yuqi-zhang yes, we could do that, in the default case it won't change how they work.  Another change we will have to do for all containers is to specify the additional systemd `ExecStartPre=$EXEC_STARTPRE` and `ExecStopPost=$EXEC_STOPPOST` so that we can finally use `runc -d` instead of requiring runc running for each container.

@ingvagabund I confirm that it works with the Openshift installer, all I needed to do, was:

1) `--set=ADDTL_MOUNTS=',{"type":"bind","source":"/etc/","destination":"/etc/","options":["rbind","rw","rslave"]}'`
2) replace any `/var/lib/etcd/etc` with `/etc/etcd`
3) Specify `DATADIR=/var/lib/etcd`

giuseppe:
@aveshagarwal how easy would be to get this change in the official etcd image (once it gets accepted)?  It is needed to drop a huge hack we are carrying in the OpenShift installer for using the etcd system container

yuqi-zhang:
@giuseppe sounds good. We can do that in a separate PR. I can merge this once the discussion is finished.

aveshagarwal:
@giuseppe create a bz for 7.4.1 so that I can commit these changes (config.json.template and manifest.json) in the next build.

ashcrow:
@rh-atomic-bot r+ e62b9ec

PR #107, opened 2017-08-25
Drop /usr/share/rhel bind mount

giuseppe:
The changes introduced with https://github.com/projectatomic/atomic-system-containers/pull/97 bring some more issues on Atomic Host where `/usr` is immutable.  To avoid any similar issue, drop completely the mount point and expect `/usr/share/rhel` to be present in the container (if needed)

### DIFF
--- a/cri-o-centos/config.json.template
+++ b/cri-o-centos/config.json.template
@@ -309,17 +309,6 @@
         },
 	{
 	    "type": "bind",
-	    "source": "/usr/share/rhel",
-	    "destination": "/usr/share/rhel",
-	    "options": [
-		"rprivate",
-		"rbind",
-		"ro",
-		"mode=755"
-	    ]
-	},
-	{
-	    "type": "bind",
 	    "source": "${RUN_DIRECTORY}",
 	    "destination": "/run",
 	    "options": [
--- a/cri-o-centos/tmpfiles.template
+++ b/cri-o-centos/tmpfiles.template
@@ -1,5 +1,4 @@
 d    ${RUN_DIRECTORY}/${NAME}               -        -           -       - -
 d    /etc/crio - - - - -
 Z    /etc/crio - - - - -
-d    /usr/share/rhel - - - - -
 d    ${STATE_DIRECTORY}/origin               -        -           -       - -
--- a/cri-o-fedora/config.json.template
+++ b/cri-o-fedora/config.json.template
@@ -314,17 +314,6 @@
         },
 	{
 	    "type": "bind",
-	    "source": "/usr/share/rhel",
-	    "destination": "/usr/share/rhel",
-	    "options": [
-		"rprivate",
-		"rbind",
-		"ro",
-		"mode=755"
-	    ]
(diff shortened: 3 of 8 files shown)


ashcrow:
In both cri-o and docker cases ``/usr/share/rhel`` is not present in the container. I think this needs to be tested before we merge.

giuseppe:
`/usr/`share/rhel` is used only the Docker system container, since we have a specialized version for the Docker system container now on RHEL I think it is fine that the docker-centos version doesn't address RHEL as well.  AFAICS, cri-o doesn't use `/usr/share/rhel` at all.

rhatdan:
The only use of this would be if we wanted to support yum install inside of the container.  Currently we don't do the equivalent of the secrets patch in CRI-O, but we could add that in the future if needed.

giuseppe:
this is not going to change anything in practice since the docker rpm on RHEL creates `/usr/share/rhel` which is populated with symlinks to files under `/etc`.  Since `/etc` is in any case coming from the host it will just work as it does now.

IIRC I think this was added for running on RHEL the Docker system container based on CentOS (where `/usr/share/rhel` is missing in the image).  Since now we have a version based on RHEL of Docker we don't really need to support this case.

If it turns out we want to support it, probably it will be better to create these files either as part of the `Dockerfile` or in the docker rpm itself (as the RHEL one does).

ashcrow:
Is there a way we can comment the config file so it doesn't get removed in downstream RHEL builds? We just saw this happen with the Openshift node image for system containers and I'd like to try to avoid it of possible.

giuseppe:
it is fine if this gets dropped in RHEL as well, since these files are coming from the docker rpm included into the system container.  This was useful only when using the CentOS based system container on RHEL.

ashcrow:
@giuseppe OK, that's fair. At some point we still need to break dependence on OS flavor when it comes to system containers but that is out of scope for this PR.

👍

ashcrow:
I'm going to do a smoke test before merging.

giuseppe:
this is what I have on RHEL with container-engine once I drop the `/usr/share/rhel` mount point:

```
# grep -c /usr/share /var/lib/containers/atomic/container-engine/config.json 
0

# runc exec container-engine ls -l /usr/share/rhel/secrets/
total 0
lrwxrwxrwx. 2 root root 20 Jul 15 11:48 etc-pki-entitlement -> /etc/pki/entitlement
lrwxrwxrwx. 2 root root 28 Jul 15 11:48 rhel7.repo -> /etc/yum.repos.d/redhat.repo
lrwxrwxrwx. 2 root root  9 Jul 15 11:48 rhsm -> /etc/rhsm
```

ashcrow:
@giuseppe I'll ensure I have the same result once the install finishes.

ashcrow:
Worked like a charm 👍

ashcrow:
@rh-atomic-bot r+ 5bac895e1a3a7d048df32b0ee788ebd3676bd9b7

PR #108, opened 2017-08-28
Remove cri-o-cni package

yuqi-zhang:
The package was dropped and now causes errors.


giuseppe:
@rh-atomic-bot r+ 60c641c

ashcrow:
Where is the cni coming from now?

giuseppe:
it is part of the crio package, it was already the case on CentOS, now the two Dockerfiles are more similar

PR #109, opened 2017-09-01
Fix SELinux label for system containers

giuseppe:
the SELinux runtime label was moved to `process/selinuxLabel`.

Drop it from system containers that are not using it.

Update the CRI-O system container to correctly use the new label.

### DIFF
--- a/lint/syscontainers-lint
+++ b/lint/syscontainers-lint
@@ -154,6 +154,8 @@ def check_config_json(path):
             log(basename, "Readonly must be true", is_error=True)
     if config['process']['terminal']:
             log(basename, "process/terminal must be false", is_error=True)
+    if 'selinuxProcessLabel' in config['linux']:
+            log(basename, "linux/selinuxProcessLabel not valid.  Use process/selinuxLabel", is_error=True)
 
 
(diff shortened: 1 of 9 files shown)


lint/syscontainers-lint
```diff
@@ -154,6 +154,8 @@ def check_config_json(path):
             log(basename, "Readonly must be true", is_error=True)
     if config['process']['terminal']:
             log(basename, "process/terminal must be false", is_error=True)
+    if 'selinuxProcessLabel' in config['linux']:
```
ashcrow: 👍 that bit us a few times.

yuqi-zhang:
@rh-atomic-bot r+ 5106d65

In the future it might be better to have less commits if most of them are just one-liners ;)

giuseppe:
@yuqi-zhang I know it looks a bit weird but I'd still suggest to have different commits when they touch different system containers.  Each system container is really a project on its own with its history completely unrelated to the other ones.  One advantage is that it is easier to cherry-pick commits if we will ever need it.

yuqi-zhang:
@giuseppe ah I see, makes sense, I will watch for that in the future.

PR #112, opened 2017-09-11
cri-o-centos: change repository to virt7-container-common-candidate

giuseppe:

### DIFF
--- a/cri-o-centos/Dockerfile
+++ b/cri-o-centos/Dockerfile
@@ -11,7 +11,7 @@ LABEL com.redhat.component="cri-o" \
       maintainer="Yu Qi Zhang <jzehrarnyg@gmail.com>" \
       atomic.type="system"
 
-RUN yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-common-candidate/x86_64/os/ && \
+RUN yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-container-common-candidate/x86_64/os/ && \
     yum install --nogpgcheck --setopt=tsflags=nodocs -y iptables cri-o iproute runc && \
     rpm -V iptables cri-o iproute runc && \
     yum clean all && \


cri-o-centos/Dockerfile
```diff
@@ -11,7 +11,7 @@ LABEL com.redhat.component="cri-o" \
       maintainer="Yu Qi Zhang <jzehrarnyg@gmail.com>" \
       atomic.type="system"
 
-RUN yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-docker-common-candidate/x86_64/os/ && \
+RUN yum-config-manager --nogpgcheck --add-repo https://cbs.centos.org/repos/virt7-container-common-candidate/x86_64/os/ && \
```
lsm5: while this should be fine moving forward, CentOS policy is kinda to discourage CBS links in favor of links to buildlogs.centos.org (testing packages) and mirror.centos.org (release packages). The packages I build on CBS are moved right away to buildlogs.centos.org (See: https://buildlogs.centos.org/centos/7/virt/x86_64/container/ )

lsm5: See: https://github.com/CentOS-virt7/centos-release-container/blob/master/CentOS-Container.repo.x86_64 for the yum repo source that would be setup via the 'centos-release-container' package.

lsm5: So, soon-ish you should be able to just 'yum install centos-release-container' and be done with it

giuseppe: I've pushed a new version but when I try that repo I get:

```
Error: Package: cri-o-1.0.0.beta.0-1.git66d96e7.el7.x86_64 (buildlogs.centos.org_centos_7_virt_x86_64_container_)
           Requires: containernetworking-plugins >= 0.5.2-2
 You could try using --skip-broken to work around the problem
```

lsm5: cbs link LGTM for now, since the buildlogs link evidently doesn't work for this case. Let me get back on the centos-release-container package is fixed and published

giuseppe:
@rh-atomic-bot delegate=lsm5

giuseppe:
@rh-atomic-bot r=ashcrow 132ebb7

PR #117, opened 2017-09-22
cri-o, docker: rbind /var/lib

giuseppe:
so we don't shadow /var/lib/docker when it is already mounted.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1493714


ashcrow:
@rh-atomic-bot r+ 9690159

PR #120, opened 2017-09-27
drop /var/lib bind mount completely in the host

giuseppe:
Having a rbind bind mount on /var/lib is causing all sort of issues in the host (e.g. ostree storage on two different file systems), so drop it completely and in case we will need to create mounts in the host under /var/lib, we will deal with that more specifically.

Marked as WIP for now, as I am still testing this change

### DIFF
--- a/cri-o-centos/config.json.template
+++ b/cri-o-centos/config.json.template
@@ -350,13 +350,13 @@
             "type": "bind"
         },
         {
-            "destination": "/var/lib/containers",
+            "destination": "/var/lib/containers/storage",
             "options": [
                 "rbind",
                 "rshared",
                 "rw"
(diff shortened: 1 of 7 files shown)


giuseppe:
I managed to setup a CRI-O cluster with this change (which is already included into docker.io/gscrivano/cri-o-centos) so I am dropping the WIP label.  It also fixes the https://bugzilla.redhat.com/show_bug.cgi?id=1495588 bug

ashcrow:
LGTM. Ready to merge?

giuseppe:
@ashcrow yes!

ashcrow:
@rh-atomic-bot r+ e219ec6

PR #121, opened 2017-09-27
Create cri-o storage directory

ashcrow:
When testing I noticed a failure on Fedora 26 as ``/var/lib/containers/storage`` didn't already exist. These commits add ``/var/lib/containers/storage`` to ``/exports/hostfs``.


giuseppe:
@rh-atomic-bot r+ 7d010df

PR #123, opened 2017-09-29
Add overlay override check

ashcrow:


ashcrow:
@rh-atomic-bot r+ fde6a10

PR #124, opened 2017-10-09
Enable mount propagation in /var/lib/kubelet

rochaporto:
Cases where an external volume driver manages the bind mounts to be made
available to containers in a different namespace need mount propagation
on /var/lib/kubelet. An example would be a driver relying on FlexVolume
which is doing the pod bind mount to a shared filesystem.

### DIFF
--- a/kubernetes-kubelet/config.json.template
+++ b/kubernetes-kubelet/config.json.template
@@ -353,7 +353,7 @@
             "destination": "/var/lib/kubelet",
             "options": [
                 "rbind",
-                "rslave",
+                "rshared",
                 "rw",
                 "mode=755"
              ]


giuseppe:
bot, test pull request

kubernetes-kubelet/config.json.template
```diff
@@ -370,7 +370,7 @@
           }
     ],
     "linux": {
-        "rootfsPropagation": "rslave",
+        "rootfsPropagation": "rshared",
```
giuseppe: wouldn't the /var/lib/kubelet rshared be enough?  If it is not sufficient, as it is not a mount point on the host, we will need to ensure it is.

This is what we do, for example, with the CRI-O system container:

https://github.com/projectatomic/atomic-system-containers/blob/master/cri-o-centos/set_mounts.sh
https://github.com/projectatomic/atomic-system-containers/blob/master/cri-o-centos/Dockerfile#L26
https://github.com/projectatomic/atomic-system-containers/blob/master/cri-o-centos/service.template#L7

The risk of having rootfsPropagation=rshared is that mount events can get out of control, and you will end up with a lot of undesired mounts in the host mount namespace.

rochaporto: I think you're right, i'll rerun some tests and let you know.

ashcrow:
bot, test pull request

jasonbrooks:
@rochaporto Do you have a test case for this? 

This week I've been trying to get to the bottom of a similar issue, using rook storage w/ system containerized kubelet. Rook is kind of picky about its install sequence, I'd like to try w/ a simpler test.

giuseppe:
LGTM, but I'll leave the decision to @jasonbrooks for the approval.

I may need to do the same change in the OpenShift Origin system container

jasonbrooks:
@rh-atomic-bot r+ ab38f66

giuseppe:
@rh-atomic-bot r=jasonbrooks ab38f66

PR #125, opened 2017-10-09
Update to current standards

ashcrow:
- Replaces ``MAINTAINER`` with the ``maintainer`` label.
- ``yum``/``dnf`` now using ``--setopt=tsflags=nodocs``

**Note**: There are some other warnings that still are printed out

- ``found mount point /var/lib.  Use ${STATE_DIRECTORY} instead``
- ``found mount point /run.  Use ${RUN_DIRECTORY} instead``

Should I move these over to use the variables or should we keep with the hard coding?

### DIFF
--- a/azure/Dockerfile
+++ b/azure/Dockerfile
@@ -1,5 +1,5 @@
 FROM centos:latest
-RUN yum -y install WALinuxAgent && yum -y clean all
+RUN yum -y --setopt=tsflags=nodocs install WALinuxAgent && yum -y clean all
 COPY exports/service.template /exports/
 LABEL Name="azure" \
       Version="0.1" \
--- a/hello-world/Dockerfile
+++ b/hello-world/Dockerfile
@@ -1,5 +1,5 @@
 FROM fedora
-RUN dnf install -y nmap-ncat && dnf clean all
+RUN dnf install --setopt=tsflags=nodocs -y nmap-ncat && dnf clean all
 
 COPY run.sh greet.sh /usr/bin/
 COPY manifest.json service.template config.json.template /exports/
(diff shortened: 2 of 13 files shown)


giuseppe:
LGTM.

About the warnings: these were supposed for images that can run both as root and as non-root. user.  If an image cannot be used by a non root user then I think it is fine to use directly /var/lib.  Since only a small subset of the images can be used by a non root user though.

I wonder if we should make it explicit in the manifest.json file, something like `userMode: true`, so that atomic can fail with a nice message if a non root user tries to install an image that is not supported.

ashcrow:
> I wonder if we should make it explicit in the manifest.json file, something like userMode: true, so that atomic can fail with a nice message if a non root user tries to install an image that is not supported.

I think that makes a lot of sense. First step would be to have the lint tool understand the difference, then, after a period of time, we can implement the check in the ``atomic`` command to fail if the image is used in a non expected way.

ashcrow:
Though having a ``userMode`` field in ``manifest.json`` will probably require ``manifest.json`` to always exist. Otherwise we'd be assuming that the image would always be system and user compatible unless the file and field are provided. I worry that people will be people and default to not providing a manifest file until they get a bug, and even then they may fall back on "don't do that" rather than having the tools guide them. @giuseppe WDYT?

giuseppe:
I think the default (or when the manifest.json file is missing) should be that the user mode is not supported so that we can keep compatibility with current system containers.  The only container I've really tried in user mode is etcd, and even that depends on bwrap-oci upstream, I need to make a new release now that bubblewrap was released.  So I think it is safe to assume there are 0 users of non root containers right now :)

giuseppe:
@ashcrow if you are fine with this PR (so no WIP), I can merge it

ashcrow:
@giuseppe dropped WIP. I'll probably do a follow on later today with some other enhancements around what we've spoken about.

giuseppe:
@rh-atomic-bot r+ a634fa3

PR #126, opened 2017-10-10
Follow on to lint fixes

ashcrow:
- Warn on missing files
- Error when ``config.json`` is used
- No longer throw stack trace is ``config.json``/``config.json.template`` is malformed
- Work on Python 2 and 3


giuseppe:
@rh-atomic-bot r+ 61f1fbe

PR #130, opened 2017-10-17
CRI-O: rbind storage path

giuseppe:
Use rbind so we don't cover any existing mount point on /var/lib/containers/storage/overlay.    We  observed this issue when Skopeo was used to copy images to the  overlay storage before CRI-O was running.  Skopeo created a bind mount  on /var/lib/containers/storage/overlay that was not accessible anymore once we covered it with /var/lib/containers/storage.

CRI-O would think /var/lib/containers/storage/overlay by parsing /proc/self/mountinfo would believe the mount point is already there and thus failing to make it PRIVATE.

### DIFF
--- a/cri-o-centos/set_mounts.sh
+++ b/cri-o-centos/set_mounts.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-findmnt /var/lib/containers/storage > /dev/null || mount --bind --make-shared /var/lib/containers/storage /var/lib/containers/storage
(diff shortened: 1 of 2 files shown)


ashcrow:
cc @jwhonce

ashcrow:
@rh-atomic-bot r+ 3404e01

giuseppe:
@ashcrow thanks for the quick review.  I've rebuilt the images now

PR #131, opened 2017-10-18
bind mount /var/lib/kubelet

jasonbrooks:
I needed to make these changes to use this system container with kubeadm.


ashcrow:
@rh-atomic-bot r+ e51f6ec

PR #134, opened 2017-10-23
fix uid/gid, bind mount /var/run/kubernetes in apiserver

### LINKED ISSUE #133: kubernetes-apiserver needs a writable directory for certificates
If the user doesn't specify any certificates, the apiserver will try to create those in /var/run/kubernetes or in the directory that is specified with --cert-dir. Currently, all dirs mounted in the apiserver container are read-only.

So, we need to mount /var/run/kubernetes as rw

jasonbrooks:
kube-apiserver needs to write a cert to /var/run/kubernetes if another cert dir isn't provided. 

atomic hosts come with the kube user set to 996 and group to 994, and I had these reversed

fixes #133

### DIFF
--- a/kubernetes-apiserver/config.json.template
+++ b/kubernetes-apiserver/config.json.template
@@ -7,8 +7,8 @@
     "process": {
         "terminal": false,
         "user": {
-            "uid": 994,
-            "gid": 996
+            "uid": 996,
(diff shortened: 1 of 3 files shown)


ashcrow:
@rh-atomic-bot r+ 9c1cda8

PR #135, opened 2017-10-23
add dir for kube

jasonbrooks:
apply change from #131 to fedora version of crio container


ashcrow:
@rh-atomic-bot r+ ae0ac9f

PR #144, opened 2017-11-14
Update to fedora:27

giuseppe:
And enable testing for CRI-O so the last version available is used.

### DIFF
--- a/cri-o-fedora/Dockerfile
+++ b/cri-o-fedora/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:26
+FROM registry.fedoraproject.org/fedora:27
 
 ENV VERSION=0 RELEASE=1 ARCH=x86_64
 LABEL com.redhat.component="cri-o" \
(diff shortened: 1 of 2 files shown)


cri-o-fedora/Dockerfile
```diff
@@ -13,7 +13,7 @@ LABEL com.redhat.component="cri-o" \
 
 COPY README.md /
 
-RUN dnf install --setopt=tsflags=nodocs -y iptables cri-o iproute runc && \
+RUN dnf install --enablerepo=updates-testing --setopt=tsflags=nodocs -y iptables cri-o iproute runc && \
```
ashcrow: Why enable testing?

ashcrow: Oh I see, it's temporary to get the latest stable ``cri-o``. Could this be done on it's own line to avoid getting testing versions of the other packages?

giuseppe: some of the packages are updated in Fedora, some others are not.  I always end up with build issues when a new release is out.  This was the suggestion I got from: @lsm5

ashcrow: OK, let's just make sure we revert back to stable once the package has moved.

ashcrow:
@rh-atomic-bot r+ 4884109

PR #149, opened 2017-11-20
CRI-O: use 0 instead of infinity

giuseppe:
The previous PR breaks the system container on RHEL, so...


runcom:
LGTM

giuseppe:
@rh-atomic-bot r=runcom f12be9d

PR #154, opened 2017-12-04
docker-centos: use directly dockerd instead of docker daemon

giuseppe:
We avoid to spawn a new process that gets a different SELinux label.


ashcrow:
Makes sense. Does this need to be in the Fedora image as well?

ashcrow:
@rh-atomic-bot r+ d7f2dff

PR #158, opened 2017-12-19
run kube-proxy as root

jasonbrooks:
The 996/994 uid/gid are for the kube user and group as they are configured
in atomic hosts. When installed as rpms, the master components run as
the kube user, so I set these ids for those system containers to work that way,
too. Following that convention, though, the proxy should be running as
root, since that's how it runs when installed via RPM.

### DIFF
--- a/kubernetes-proxy/config.json.template
+++ b/kubernetes-proxy/config.json.template
@@ -6,7 +6,10 @@
     },
     "process": {
         "terminal": false,
-        "user": {},
+        "user": {
+            "uid": 0,
+            "gid": 0
+        },
         "args": [
             "/usr/bin/kube-proxy-docker.sh"
         ],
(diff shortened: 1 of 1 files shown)


kubernetes-proxy/config.json.template
```diff
@@ -6,10 +6,7 @@
     },
     "process": {
         "terminal": false,
-        "user": {
-            "uid": 994,
-            "gid": 996
-        },
+        "user": {},
```
ashcrow: Other system images denote:

```json
"user": {
    "uid": 0,
    "gid": 0
},
```

Is leaving this empty fine?

jasonbrooks: Leaving it out defaults to 0, but it's probably clearer to be explicit. I'll change it.

ashcrow:
@rh-atomic-bot r+ c328cde

ashcrow:
@rh-atomic-bot r+ a72c882

Issue #155, opened 2017-12-10
kube-proxy fails to expose NodePort because of r/o filesystem


neuhalje:
The containerised kube-proxy fails to expose services with `NodePort` because it cannot lock `/run/xtables.lock`  (`open /run/xtables.lock: read-only file system`).
.

## Version used

```text
sudo atomic images list
...
>  registry.fedoraproject.org/f27/kubernetes-proxy     latest   68406693c322   2017-12-10 17:26   237.16 MB      ostree
```

## Service definition

Given the following yaml:

```yaml
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.8
        ports:
        - containerPort: 80

---
kind: Service
apiVersion: v1
metadata:
  name: my-nginx
spec:
  type: NodePort
  selector:
    app: nginx
  ports:
  - protocol: TCP
    port: 80
    targetPort: 80
```

### kubectl

```text
kubectl get pods

NAME                                READY     STATUS    RESTARTS   AGE
nginx-deployment-3718365652-1ht6q   1/1       Running   1          1h
nginx-deployment-3718365652-9g2jq   1/1       Running   1          1h
```

```text
kubectl get service my-nginx

NAME       TYPE       CLUSTER-IP      EXTERNAL-IP   PORT(S)        AGE
my-nginx   NodePort   10.254.106.83   <none>        80:32315/TCP   1h
```

## Expected behaviour

* `curl http://172.20.61.51:32315` should return the nginx page.

## Observed behaviour

The port is not exposed.

```text
sudo netstat -tulpen | grep proxy

tcp        0      0 127.0.0.1:10249         0.0.0.0:*               LISTEN      994        25072      797/kube-proxy
tcp6       0      0 :::10256                :::*                    LISTEN      994        24025      797/kube-proxy
```
Although I can connect to the ports of the container:

```text
curl http://127.0.0.1:10249/

404 page not found
```

`journaltctl -xe -u kube-proxy.service` returns the following errors:

```text
Failed to start in resource-only container "/kube-proxy": mkdir /sys/fs/cgroup/cpuset/kube-proxy: read-only file system
...
[...]

PR #161, opened 2018-02-08
Update kubelet based openshift/node

strigazi:
Mount /sys/ as in node to avoid kubelet log:
Failed to get system container stats for
"/systemd/system.slice": failed to get cgroup stats
for "/systemd/system.slice": failed to get container
info for "/systemd/system.slice": unknown container
"/systemd/system.slice".

Mount /etc/pki to use the ca from the host.

Mount /var/run/secrets needed for f27 which doesn't have
kube installed and it is required by kubelet to manage
kubernetes secrets.

Mount tmp as tmpfs.

Mount cni dirs.

### DIFF
--- a/kubernetes-kubelet/tmpfiles.template
+++ b/kubernetes-kubelet/tmpfiles.template
@@ -1,0 +1,2 @@
+d   /var/lib/cni - - - - -
+d   /var/run/secrets - - - - -
--- a/kubernetes-kubelet/Dockerfile
+++ b/kubernetes-kubelet/Dockerfile
@@ -18,8 +18,10 @@ LABEL RUN /usr/bin/docker run -d --privileged --net=host --pid=host -v /:/rootfs
 
 COPY launch.sh /usr/bin/kubelet-docker.sh
 
-COPY service.template config.json.template /exports/
+COPY tmpfiles.template service.template config.json.template /exports/
 
-RUN mkdir -p /exports/hostfs/etc/kubernetes && cp /etc/kubernetes/{config,kubelet} /exports/hostfs/etc/kubernetes
+RUN mkdir -p /exports/hostfs/etc/cni/net.d && \
+    mkdir -p /exports/hostfs/etc/kubernetes && \
+    cp /etc/kubernetes/{config,kubelet} /exports/hostfs/etc/kubernetes
 
 ENTRYPOINT ["/usr/bin/kubelet-docker.sh"]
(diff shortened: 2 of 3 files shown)


strigazi:
Based on https://github.com/openshift/origin/tree/master/images/node/system-container

strigazi:
@jasonbrooks The error that I mention in the commit message and I posted on the channel yesterday is fixed by mounting /sys/ like in openshift/node

ashcrow:
bot, test pull request

strigazi:
fixes #155 too

strigazi:
I addressed this error:
Unknown mount type rbind.  Use bind

I don't understand where it fails now. btw in openshift they have type rbind

strigazi:
should be ok now

strigazi:
@giuseppe can you check this mount if it's ok?
https://github.com/projectatomic/atomic-system-containers/pull/161/commits/eea95b44f84067ca561e1ee5a52e8ae12093f694#diff-4b500d991e0f4b2604ba76c6f5a63351R269

versus this on openshift
https://github.com/openshift/origin/blob/master/images/node/system-container/config.json.template#L277

kubernetes-kubelet/tmpfiles.template
```diff
@@ -0,0 +1,3 @@
+d   /etc/cni/net.d - - - - -
```
giuseppe: do you want this to be a temporary file?  systemd will destroy/recreate it on reboot.

If you want it to be persistent, you could do:
```
RUN mkdir -p /exports/hostfs/etc/cni/net.d
```
in your Dockerfile

giuseppe:
The mounts looks fine to me, just I have one inline comment

kubernetes-kubelet/tmpfiles.template
strigazi: I created the dir in the dockerfile, do  I also need to remove it from this list?

giuseppe: yes please.  Otherwise systemd-tmpfiles will still try to handle it.

strigazi:
@jasonbrooks @ashcrow @giuseppe let's take this in?

jasonbrooks:
@strigazi make that tmpfiles.template change and I'll approve

strigazi:
done

jasonbrooks:
@rh-atomic-bot r+ 88c81b7

PR #162, opened 2018-02-16
buildah-fedora: new system container

giuseppe:
Blocked on: https://github.com/projectatomic/atomic/pull/1186

### DIFF
--- a/buildah-fedora/Dockerfile
+++ b/buildah-fedora/Dockerfile
@@ -1,0 +1,15 @@
+FROM registry.fedoraproject.org/fedora:27
+
+ENV VERSION=0 RELEASE=1 ARCH=x86_64
+LABEL com.redhat.component="buildah" \
+      atomic.run="once" \
+      atomic.type="system"
+
+RUN mkdir /exports /host && \
+    dnf install --setopt=tsflags=nodocs -y buildah && \
+    dnf clean all
+
+# system container
+COPY config.json.template /exports/
+
+CMD ["/usr/bin/false"]
--- a/buildah-fedora/README.md
+++ b/buildah-fedora/README.md
@@ -1,0 +1,19 @@
+# buildah-container
+
+## Running the system container
+
+The Buildah system container doesn't install a service, but it requires
+`atomic run`.  The container has a mount point `/host` to the host.
+
+It is possible to run it as:
+
+```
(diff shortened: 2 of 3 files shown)


giuseppe:
I experimented with it as part of a comment here: https://github.com/projectatomic/buildah/issues/158

The nice part is that it works as an user container as well, bwrap-oci can run it.

Opening a PR so it doesn't get lost

buildah-fedora/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM registry.fedoraproject.org/fedora:27
+
+ENV VERSION=0 RELEASE=1 ARCH=x86_64
+LABEL com.redhat.component="buildah" \
+      atomic.run="once" \
+      atomic.type="system"
+
+RUN mkdir /exports /host && dnf install --setopt=tsflags=nodocs -y buildah && dnf clean all
```
ashcrow: nit: It's always nice to have commands on different lines.

rhatdan:
@alexlarsson PTAL

buildah-fedora/Dockerfile
giuseppe: thanks, I've pushed a new version to split this on multiple lines

giuseppe:
no, we are not able to copy files to the host with `run.once` but if we decide to go this way, we can take advantage of `noContainerService`.

With `noContainerService` we can have a container installed that doesn't require any running systemd service and we can use it to exec into it from the host.

ashcrow:
@giuseppe I'd love that! ❤️

alexlarsson:
Having this rinnande without root is awesome, but Using the vfs backend is kinda lame. Did anyone look at making a backend using the ostree rofiles-fuse helper? Especially with the new copy-up feature. This would be like an non-root overlayfs.

rhatdan:
Not yet we would have to get that added as a storage driver for containers/storage it it would work @giuseppe WDYT?

giuseppe:
@rhatdan I think that is a great idea, I have a PR for adding ostree as option for overlay, but that is completely different than this use case.

We can followup on containers/storage for the new storage driver

giuseppe:
projectatomic/atomic#1186 is approved.  Fine to :ship: this?

ashcrow:
@rh-atomic-bot r+ bbcc66b

PR #164, opened 2018-02-21
buildah: few fixes

giuseppe:
a couple of fixes in the Buildah system container:

1) Set CMD/ENTRYPOINT so it can be used as a Docker container as well
2) Mount `/etc` from the host so to inherit all the configuration in the container


ashcrow:
@rh-atomic-bot r+ c214afa0d4d85c520286d966c2357bb1fe98b9f8

PR #165, opened 2018-02-21
kubelet: Add ${STATE_DIR}/kubelet to tmpfiles

strigazi:
In f27 /var/lib/kubelet doesn't exists, we need
to create it.


strigazi:
@giuseppe Is this the right approach like in openshift/node or create it in the dockerfile like this one:
https://github.com/projectatomic/atomic-system-containers/blob/master/kubernetes-kubelet/Dockerfile#L24

giuseppe:
I think this is fine as anyway systemd will not delete the directory if it is not empty.

jasonbrooks:
@rh-atomic-bot r+ 0b4e806

PR #166, opened 2018-02-22
buildah: bind mount `/var/run/NetworkManager`

giuseppe:
bind mount `/var/run/NetworkManager` from the host so that even in the rootless case `/etc/resolv.conf` points to the correct file.

Also, add `bwrap-oci` inside the container for using it as an alternative OCI runtime


giuseppe:
@rh-atomic-bot r+

PR #167, opened 2018-02-24
docker: add support for additional mounts

giuseppe:
use the same pattern used for the `etcd` system container.

Additional mounts can be configured with `--set` at installation time

### DIFF
--- a/docker-centos/Dockerfile
+++ b/docker-centos/Dockerfile
@@ -19,7 +19,7 @@ yum clean all
 ADD init.sh /usr/bin
 
 # system container
-COPY service.template tmpfiles.template config.json.template /exports/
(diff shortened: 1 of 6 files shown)


ashcrow:
Side note: Maybe we should add ``ADDTL_MOUNTS`` as part of the ``config.json.template`` by default when using ``system-buildah``.

ashcrow:
@rh-atomic-bot r+ b903358

ashcrow:
@rh-atomic-bot r+ e3a6051

giuseppe:
@jwhonce FYI

ashcrow:
@jwhonce I don't believe so. This is an added feature for Openshift to use.

PR #168, opened 2018-03-08
etcd: make sure /etcd is owned by HOST_UID/HOST_GID

vrutkovs:
Atomic resets /var/lib/etcd to be owned by etcd:etcd, which breaks
etcd system container after reboot

Related to https://github.com/openshift/openshift-ansible/pull/7443
See https://bugzilla.redhat.com/show_bug.cgi?id=1553084


ashcrow:
@rh-atomic-bot r+ 1f6dc79

Issue #137, opened 2017-10-27
Runtime spec changed, config.json files are not usable anymore by runc


bagasse:
I tried to use the `ovirt-guest-agent-centos` system container (thanks @giuseppe) . Install works fine (followed https://www.projectatomic.io/blog/2016/09/intro-to-system-containers/)
But the agent failed to run. When I try to run it by hand to know why, I've an `json: cannot unmarshal array into Go struct field Process.capabilities of type specs.LinuxCapabilities` error from `runc`.

From https://github.com/opencontainers/runtime-spec/blob/master/config.md, the runtime spec changed and capabilities have to be like

```
"capabilities": {
        "bounding": [
                "CAP_AUDIT_WRITE",
                "CAP_KILL",
                "CAP_NET_BIND_SERVICE"
        ],
        "effective": [
                "CAP_AUDIT_WRITE",
                "CAP_KILL",
                "CAP_NET_BIND_SERVICE"
        ],
        "inheritable": [
                "CAP_AUDIT_WRITE",
                "CAP_KILL",
                "CAP_NET_BIND_SERVICE"
        ],
        "permitted": [
                "CAP_AUDIT_WRITE",
                "CAP_KILL",
                "CAP_NET_BIND_SERVICE"
        ],
        "ambient": [
                "CAP_AUDIT_WRITE",
                "CAP_KILL",
                "CAP_NET_BIND_SERVICE"
        ]
},
```

Tested on CentOS Atomic host:
```
bash-4.2# atomic host status
State: idle
Deployments:
● centos-atomic-host:centos-atomic-host/7/x86_64/standard
                Version: 7.1708 (2017-09-15 15:32:30)
                 Commit: 33b4f0442242a06096ffeffadcd9655905a41fbd11f36cd6f33ee0d974fdb2a8
           GPGSignature: 1 signature
                         Signature made ven. 15 sept. 2017 19:17:39 CEST using RSA key ID F17E745691BA8335
                         Good signature from "CentOS Atomic SIG <security@centos.org>"

  centos-atomic-host:centos-atomic-host/7/x86_64/standard
                Version: 7.1705.1 (2017-06-20 18:46:11)
                 Commit: 550d13d3e1fda491afab6d368adc13e307512ea51734055d83f0cc4d9049e91d
           GPGSignature: 1 signature
                         Signature made mar. 20 juin 2017 20:47:45 CEST using RSA key ID F17E745691BA8335
[...]

PR #176, opened 2019-08-11
Fix: qemu-guest-agent/config.json.template

MartinK07:
Same as https://github.com/projectatomic/atomic-system-containers/issues/137


davidbrowning (approved):
This seems to work

